PR #10, opened 2017-03-31
Fix windows

mattn:
fixing build error on msys2/mingw64.


taku910:
Thank you for the fix.

PR #12, opened 2017-05-15
added initial support to compile on OSX

kashif:


taku910:
Thank you for the patch.

PR #13, opened 2017-05-16
Add support to use self-prepared protobuf lib

resec:
I am trying to build sentencepiece on iOS along with tensorflow. And I want to build sentencepiece using tensorflow's costumed protobuf. By adding a env-var checking to configure.ac, It worked. I also add some guide line to README.md on how to use self-prepared protobuf lib.


taku910:
Thank you for the patch.

PR #17, opened 2017-06-14
spm_encode should not ignore empty lines

tsuchm:
The current version of spm_encode ignores empty lines, thus the line number of its input file does not match the line number of its output file.   The filter program such as spm_encode should do its most simple work only.  If its user wants to remove empty lines, he/she can do it with `grep' command as you know.


tsuchm:
I signed CLA.

PR #19, opened 2017-07-20
std::istream interface for SentencepieceProcessor construction

resec:
Added two construction interface for SentencepieceProcessor, using std::istream to initialize the model.
These interfaces are very useful in case when we construct the model from Android assets or Network where we can only have the input stream of the model file.


taku910:
Thanks for the change. This looks excellent.

SentencePiece is implemented with Google C++ coding standard, which doesn't allow to use a non-const reference as an argument. Could you use pointer instead?

void Load(std::istream *is) {
   CHECK_NOTNULL(is);
   ...

https://google.github.io/styleguide/cppguide.html#Reference_Arguments

Thank you.

resec:
@taku910  Thanks and no problem, will change the code accordingly soon.

PR #21, opened 2017-08-24
SWIG wrapper to simple sentencepiece model

sahasan:
- make sentencepiece (follow README)
- make Python module:
  make -f makefile.swig
  (patch the Python version in the makefile if not 3.4)
- install _spm.so and spm.py to your PYTHONPATH

Usage:
```
>>> import spm
>>> m = spm.SentencePieceModel("test.model")
>>> print m.encode("this is a test")
▁this ▁is ▁a ▁test
>>> print m.decode(m.encode("this is a test"))
this is a test
```

### DIFF
--- a/makefile.swig
+++ b/makefile.swig
@@ -1,0 +1,19 @@
+py_ver = 3.4
+
+all: _spm.so
+
+spm_wrap.cxx: spm.i
+	swig $(SWIGFLAGS) -python -py3 -c++ spm.i
+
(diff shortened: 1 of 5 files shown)


taku910:
I've just made a python wrapper that directly wraps SentencePieceProcessor class.  EncodeAsIds and EncodeAsPieces returns the *list* of ids and pieces.

Issue #25, opened 2017-10-11
Build error of Python module for Python 3.5.3


tsuchm:
Hi,

When building the Python module of sentencepiece for Python 3.5.3, I saw the following build error "‘PyString_AsStringAndSize’ was not declared in this scope".

In the same environment, I successfully built the Python module of sentencepiece for Python 2.7.13, thus, I believe that my environment meets build dependency of the Python module of sentencepiece.

Could you give me an advice to avoid this problem?

```
$ pip3 --no-cache-dir install --user sentencepiece
Collecting sentencepiece
  Downloading sentencepiece-0.0.0.tar.gz (183kB)
    100% |████████████████████████████████| 184kB 6.2MB/s 
Installing collected packages: sentencepiece
  Running setup.py install for sentencepiece ... error
    Complete output from command /usr/bin/python3 -u -c "import setuptools, tokenize;__file__='/tmp/pip-build-yqcu2vo3/sentencepiece/setup.py';f=getattr(tokenize, 'open', open)(__file__);code=f.read().replace('\r\n', '\n');f.close();exec(compile(code, __file__, 'exec'))" install --record /tmp/pip-oqqfb0ui-record/install-record.txt --single-version-externally-managed --compile --user --prefix=:
    running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.5
    copying sentencepiece.py -> build/lib.linux-x86_64-3.5
    running build_ext
    building '_sentencepiece' extension
    creating build/temp.linux-x86_64-3.5
    x86_64-linux-gnu-gcc -pthread -DNDEBUG -g -fwrapv -O2 -Wall -Wstrict-prototypes -g -fdebug-prefix-map=/build/python3.5-MLq5fN/python3.5-3.5.3=. -fstack-protector-strong -Wformat -Werror=format-security -Wdate-time -D_FORTIFY_SOURCE=2 -fPIC -I/usr/include/python3.5m -c sentencepiece_wrap.cxx -o build/temp.linux-x86_64-3.5/sentencepiece_wrap.o -std=c++11 -g -O2 -fdebug-prefix-map=/home/tsuchiya/work/pkg-sentencepiece=. -fstack-protector-strong -Wformat -Werror=format-security
    cc1plus: warning: command line option ‘-Wstrict-prototypes’ is valid for C/ObjC but not for C++
[...]

PR #26, opened 2017-10-12
Fix the incompatibility between Python2 and Python3.

tsuchm:
This patch fixes the issue #25


taku910:
Thank you for the fix.

PR #34, opened 2017-11-01
Fix an incompatibility problem on setup.py for Python 3.

ksudoh:
Fixed a build error by Python 3.
subprocess.check_output returns 'bytes' not 'str' in Python 3.


ksudoh:
I signed the CLA

taku910:
Thank you for fixing this!

PR #48, opened 2018-03-29
Fix inconsistent default vocab size

tetsuok:
There is a discrepancy between actual default value `vocab_size` (8000) and its comment (32k). 
This should be fixed. I'm not sure which one is preferred, though.

---
This change is [Reviewable](https://reviewable.io/reviews/google/sentencepiece/48)


taku910:
Thank you for the fix

PR #50, opened 2018-03-31
Remove unused variables

tetsuok:
This PR removes unused variables in `unigram_model.cc` and `unigram_model_trainer.cc`.

---
This change is [Reviewable](https://reviewable.io/reviews/google/sentencepiece/50)


taku910:
Thank you for the fix.

PR #51, opened 2018-04-01
Provide detailed instructions for installing the protobuf library

dongjinleekr:
This update resolves (or answers) #2; There is no package named `libprotobuf-c++` or `protocolbuffer`. Instead, there are different packages per each of Ubuntu distros.


taku910:
thank you for the fix.

hidetokazawa:
It seems like this pull request broke the formatting of README.md?

dongjinleekr:
@hidetokazawa I's sorry. Here is the fix: https://github.com/google/sentencepiece/pull/52

PR #52, opened 2018-04-02
Fix broken format

dongjinleekr:


taku910:
Thanks. This issue is already fixed.

Issue #111, opened 2018-06-18
Problems installing Python bindings on Mac


neubig:
Hello! I'm having some trouble installing the Python bindings on Mac OS, and just thought I'd mention it here in case anyone had similar trouble. This is within an Anaconda environment, Python 3.

Move into directory and install:

    $ cd /path/to/sentencepiece
    $ ./autogen.sh
    $ ./configure
    $ make
    $ make install

Success. Try to install Python bindings:

    $ python setup.py build
    Package sentencepiece was not found in the pkg-config search path.
    Perhaps you should add the directory containing `sentencepiece.pc'
    to the PKG_CONFIG_PATH environment variable
    No package 'sentencepiece' found
    Failed to find sentencepiece pkgconfig

Fix the path.

    $ export PKG_CONFIG_PATH=`pwd`/..

And try again:

    $ python setup.py build
    running build
    running build_py
    running build_ext
    building '_sentencepiece' extension
    gcc -Wno-unused-result -Wsign-compare -Wunreachable-code -DNDEBUG -g -fwrapv -O3 -Wall -Wstrict-prototypes -I/anaconda3/envs/python3/include -arch x86_64 -I/anaconda3/envs/python3/include -arch x86_64 -I/anaconda3/envs/python3/include/python3.6m -c sentencepiece_wrap.cxx -o build/temp.macosx-10.7-x86_64-3.6/sentencepiece_wrap.o -std=c++11 -g -O2 -I/Users/neubig/usr/include
    In file included from sentencepiece_wrap.cxx:3124:
    /Users/neubig/usr/include/sentencepiece_processor.h:141:8: error: no template named 'unique_ptr' in namespace 'std'
      std::unique_ptr<Rep> rep_;
      ~~~~~^
    /Users/neubig/usr/include/sentencepiece_processor.h:181:8: error: no template named 'unique_ptr' in namespace 'std'
      std::unique_ptr<std::string> rep_;
      ~~~~~^    
    ...

The strange thing is that when I run the command on its own by copy-pasting things seem to work.

[...]

PR #112, opened 2018-06-18
Fixed mac compile error

neubig:
This is an attempt to fix https://github.com/google/sentencepiece/issues/111 and I've confirmed that I can install after applying this.

I'd like to explain this fix, but upon Googling I found my own post from before, which means that apparently, I've previously encountered these issues in my environment (https://github.com/neulab/xnmt/issues/199) and this was the fix I applied there as well. I had completely forgotten about doing this and I don't remember where I referenced when I fixed the original problem. I don't know if this is a problem just with my specific environment, or may be compiling within Anaconda, or something else.

Anyway, for reference, here is my environment info:

    $ gcc --version
    Configured with: --prefix=/Applications/Xcode.app/Contents/Developer/usr --with-gxx-include-dir=/usr/include/c++/4.2.1
    Apple LLVM version 9.1.0 (clang-902.0.39.2)
    Target: x86_64-apple-darwin17.5.0
    Thread model: posix
    InstalledDir: /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin

macOS version: 10.13.4

Feel free to close this if it doesn't seem useful, but I thought I'd post it here in case it was helpful to anyone else.


DeNeutoy:
I also needed this fix - thanks!

taku910:
Hi,

Could you try the latest version?  You  would like to install sentencepiece C++ library again to build the latest python wrapper.

I've also updated the travis-ci rule which covers the python build on default python(2) and python2/3 installed via brew command.  The ci is now green on xcode 9.{2,3,4}  environment. 

The "job log" of Mac build might be useful.
https://travis-ci.org/google/sentencepiece

neubig:
Thanks! The master branch is still not working in my environment, but this branch is (after merging the latest master).

PR #130, opened 2018-07-13
TF Op - Remove hardcoded paths in make wheel script and add darwin support

chiragjn:
- Removed hardcoded paths in make_py_wheel.sh
- Changed some g++ flags to allow compiling on Darwin platforms with clang
- Had to add `-fPIC` to CXXFLAGS in configure.ac to be able to link libsentencepiece.a when building the op

I haven't worked a lot on make scripts and making wheel packages, so I am all open for any improvements

Tests pass on Python 2.7.x on
- Apple LLVM version 9.1.0 (clang-902.0.39.2) - macOS 10.13
- Linux (gcc version 4.8.5) - Ubuntu 14.04.2

The built module works for python3 but tests fail because of encoding-decoding errors

Also, I don't have access to more machines, so anyone else testing would help


chiragjn:
I signed it!

taku910:
Thank you for the PR.

Since we've changed the build tool from configure to cmake, we will not be able to use this patch directly.
Instead, I made a script just for building wheel package on mac based on this PR. Thank you.

https://github.com/google/sentencepiece/blob/master/tensorflow/make_py_wheel_mac.sh
https://travis-ci.org/google/sentencepiece/jobs/416230479  (build log)

Please let me know if you have any questions and concern.
The wheel package is going to be uploaded to pypi soon.

PR #131, opened 2018-07-13
Updating install instructions for OSX/macOS

mmistele:
ldconfig is a linux command; update_dyld_shared_cache seems to work (courtesy of https://stackoverflow.com/a/45920815, verified on my computer).


mmistele:
I signed it!

PR #143, opened 2018-07-27
invalid test causing all pieces to be identified as non valid 

jsenellart:
when split_by_unicode_script() is true 

using clang 9.1:
```
warning: comparison of constant -1 with expression of type 'unicode_script::ScriptType' is always true [-Wtautological-constant-out-of-range-compare]
```


taku910:
Thank you for the fix.

PR #242, opened 2018-11-12
Typo on paragraph #44

kant:


taku910:
Thank you for the fix.

PR #417, opened 2019-10-31
Fix version extraction from library files

brenting:
tensorflow==1.15.0

As a library file does not exists for tensorflow version 1.15.0, this command is invoked. However, the regex fails to extract the correct versions, if there are other version numbers in the full path of the library files.

**Example:**
The following path:
`"/home/<user>/<project>/.venv/lib/python3.7/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0"`

Gets parsed by the regex as:
`"3.7/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0"`

And so:
`so_file = "/home/<user>/<project>/.venv/lib/python3.7/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.3.7/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0"`

Which results in:
`tensorflow.python.framework.errors_impl.NotFoundError: <path>: cannot open shared object file: No such file or directory`

The fix applies the regex only on the basename of the library file path.

### DIFF
--- a/tensorflow/tf_sentencepiece/sentencepiece_processor_ops.py
+++ b/tensorflow/tf_sentencepiece/sentencepiece_processor_ops.py
@@ -35,7 +35,7 @@ if not hasattr(tf, 'no_gradient'):
 
 if not os.path.exists(so_file):
   versions = [
-      re.search('[0-9]+\.[0-9\.]+.*$', n).group(0)
+      re.search('[0-9]+\.[0-9\.]+.*$', os.path.basename(n)).group(0)
       for n in glob.glob(so_base + '.*')
   ]
   latest = sorted(versions, key=LooseVersion)[-1]


brenting:
The same problem also arises with:
tensorflow==2.0.0
sentencepiece==0.1.83
tf-sentencepiece==0.1.83

It can be reproduced in Google Colab with:
```
%tensorflow_version 2.x
!pip3 install --quiet sentencepiece
!pip3 install --quiet tf-sentencepiece

import tensorflow as tf
import tf_sentencepiece
```

Where it results in:
`NotFoundError: /usr/local/lib/python3.6/dist-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.3.6/dist-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0-beta1: cannot open shared object file: No such file or directory`

For tensorflow 2.0.0, the bug does not appear with (tf-)sentencepiece 0.1.84, since the "-beta1" tag is ommitted from the library file in the newer version of tf-sentencepiece and the file is thus found directly.

C-K-Loan:
Also happens for TF 2.1

PR #443, opened 2020-01-16
Add vcpkg installation instructions

LilyWangL:
Sentencepiece is available as a port in vcpkg, a C++ library manager that simplifies installation for sentencepiece and other project dependencies. Documenting the install process here will help users get started by providing a single set of commands to build sentencepiece, ready to be included in their projects.

We also test whether our library ports build in various configurations (dynamic, static) on various platforms (OSX, Linux, Windows: x86, x64, UWP, ARM) to keep a wide coverage for users.

I'm a maintainer for vcpkg, and here is what the port script looks like. We try to keep the library maintained as close as possible to the original library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -119,6 +119,18 @@ To enable subword regularization, you would like to integrate SentencePiece libr
 
 ## Installation
 
+### Building sentencepiece - Using vcpkg
+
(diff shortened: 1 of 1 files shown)


LilyWangL:
> @googlebot I signed it!

Issue #456, opened 2020-02-26
Library version selection breaks for TF 2.1


jjedele:
The regex to looking for the most recent library version if no exact match is found extracts the whole match instead of the version number. This leads to the .so extension being duplicated.

Produces:
UserWarning: No so file is found for [2.1.0] from [so.1.15.0, so.1.14.0, so.2.0.0, so.1.13.1]
Should be:
UserWarning: No so file is found for [2.1.0] from [1.15.0, 1.14.0, 2.0.0, 1.13.1]

I'll fix and open a PR right away.

jjedele:
Hmm, somehow GitHub doesn't let me create a PR for this repo (error 500).

Here's the fix of somehow wants to pull or just fix it here: https://github.com/jjedele/sentencepiece/commit/4d050563cebeaee5e21f91597af5b12c61dacf07

JLHasson:
For what it's worth I am also seeing this on `tf-sentencepiece==0.1.85`.

```
File "/Users/lancehasson/miniconda3/envs/<package>/lib/python3.6/site-packages/tensorflow_core/python/framework/load_library.py", line 57, in load_op_library
    lib_handle = py_tf.TF_LoadLibrary(library_filename)
tensorflow.python.framework.errors_impl.NotFoundError: dlopen(/Users/lancehasson/miniconda3/envs/<package>/lib/python3.6/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.so.2.0.0, 6): image not found
```

Selected file should be `_sentencepiece_processor_ops.so.2.0.0` not `_sentencepiece_processor_ops.so.so.2.0.0`, which is the file that does indeed exist:

```
$ ls -lh /Users/lancehasson/miniconda3/envs/<package>/lib/python3.6/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0
-rwxr-xr-x  1 lancehasson  staff   814K Mar 20 13:39 /Users/lancehasson/miniconda3/envs/<package>/lib/python3.6/site-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0
```

@taku910 any chance we can get someone to take a look at #457?

taku910:
Fixed. Please try the latest version.

PR #457, opened 2020-02-27
Fix library selection logic

jjedele:
In the branch which is executed if no exact match is found,
extract the captured version number from the regex instead
of the complete match.

See issue #456 .

### DIFF
--- a/tensorflow/tf_sentencepiece/sentencepiece_processor_ops.py
+++ b/tensorflow/tf_sentencepiece/sentencepiece_processor_ops.py
@@ -35,7 +35,7 @@ if not hasattr(tf, 'no_gradient'):
 
 if not os.path.exists(so_file):
   versions = [
-      re.search('so.([0-9]+\.[0-9\.]+.*)$', os.path.basename(n)).group(0)
+      re.search('so.([0-9]+\.[0-9\.]+.*)$', os.path.basename(n)).group(1)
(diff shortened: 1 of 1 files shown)


jjedele:
Travis build fails, but I don't think it's related to my change:

> Overall coverage rate:
>   lines......: 95.6% (5177 of 5414 lines)
>   functions..: 90.0% (2875 of 3194 functions)
> [100%] Built target coverage
> + coveralls --exclude-pattern .*(include|usr|test|third_party|pb|_main).* --gcov-options \-lp --gcov gcov
> Traceback (most recent call last):
>   File "/usr/local/bin/coveralls", line 10, in <module>
>     sys.exit(run())
>   File "/usr/local/lib/python2.7/dist-packages/cpp_coveralls/__init__.py", line 92, in run
>     raise ValueError("\nno coveralls.io token specified and no travis job id found\n"
> ValueError: 
> no coveralls.io token specified and no travis job id found
> see --help for examples on how to specify a token
> The command "$TRAVIS_BUILD_DIR/test.sh ${IMAGE} ${COMMAND}" exited with 1.

mariokostelac:
I've bumped into the same issue. It definitely helps. I think the other issue with the logic is that it always chooses the newest version of the file, which I am not sure whether it's correct or. For example. Can TF 1.15.2 work with sentecepiece so for 2.0.0?

PR #476, opened 2020-04-17
Fix version detection bug

aboSamoor:
Currently, the code return paths with double extensions '.so' such as 
'/usr/local/lib/python3.6/dist-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.so.2.0.0'

This is result from return group 0 instead of 1 in version detection codebase, switching to group 1 fixes the issue to
'/usr/local/lib/python3.6/dist-packages/tf_sentencepiece/_sentencepiece_processor_ops.so.2.0.0'


taku910:
The same PR is already applied. Thank you.

PR #515, opened 2020-06-26
[Bugfix] Add split_digits to SentencePieceTrainer spec parser

mingruimingrui:
Got your back.


mingruimingrui:
Fixes issue where split_digits option isn't working from the Python API.

PR #527, opened 2020-08-05
Tutorial to add new tokens to trained sentencepiece model

gusalsdmlwlq:
I made a simple example to add new tokens to already trained sentencepiece model in python.
This maybe helpful for people want to use pre-trained NLP models using sentencepiece.


gusalsdmlwlq:
@googlebot I signed it!

PR #542, opened 2020-09-16
Update make_py_wheel.sh

AWSjswinney:
updating from manylinux1 to manylinux2014  as : (a) manylinux1 will be depreciated soon , and (b) need manylinux2014 to support new architecture like arm64


AWSjswinney:
@googlebot I signed it!

PR #550, opened 2020-10-01
Add option list to README

stephantul:
Hi, 

thanks for the great package. I was missing an overview of parameters to pass to `SentencePieceTrainer`. When just using `pip install sentencepiece`, `spm_train` is not installed by default, and, as such, the `spm_train --help` command is not available. This PR adds an overview of the options (basically just an md file with the output of `spm_train --help` pasted in.


stephantul:
@googlebot I signed it!

PR #570, opened 2020-10-22
fix bpe dropout with alpha == 0 (now yields deterministic BPE)

### LINKED ISSUE #569: Bug with BPE dropout?
I am attempting to use the BPE-dropout feature from the either the command line or the python API. I show the python example because `spm_encode` doesn't support dropout-bpe at all based on `spm_encode --help`.

Im using version `0.1.93`.

I start by making my BPE model. There was no mention of dropout in `spm_train --help` so I assume we don't have to specify anything here.
```
spm_train --input=spm.corpus --model_prefix=spm --vocab_size=32000 --model_type=bpe --byte_fallback
```
We will now use this model to tokenise stochastically from Python. Looking at `help(s.encode)` I read `alpha: Soothing parameter for unigram sampling, and merge probability for
      BPE-dropout.` The parameter seems to have no effect unless we set `enable_sampling=True`, for example:
```
>>> import sentencepiece as spm
>>> s = spm.SentencePieceProcessor(model_file='spm.model')
>>> s.encode('New York', out_type=str, alpha=1.0)
['▁New', '▁York']
>>> s.encode('New York', out_type=str, alpha=0.0)
['▁New', '▁York']
>>> s.encode('New York', out_type=str, alpha=0.5)
['▁New', '▁York']
```

When we do set it, I expected that `alpha=1.0` would yield character segmentation (always drop-out) while `alpha=0.0` would yield deterministic BPE. *(Actually the documentation mentions `merge probability` not `dropout probability` so I would expect the opposite but let's assume there was a misspelling in the doc)*.

In any case while `alpha = 1.0` works, `alpha = 0.0` doesn't. The function seems to be unaware that my model is BPE and not Kudo's LM based on the error message:

```
>>> # Good
>>> s.encode('New York', out_type=str, enable_sampling=True, alpha=1.0)
['▁', 'N', 'e', 'w', '▁', 'Y', 'o', 'r', 'k']
>>> # Here I expected deterministic BPE
>>> s.encode('New York', out_type=str, enable_sampling=True, alpha=0.0)
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/mnt/work/home/estergiadis/tf2.0/lib/python3.6/site-packages/sentencepiece.py", line 268, in Encode
    'When enable_sampling is True, We must specify "nbest_size > 1" or "nbest_size = -1", '
RuntimeError: When enable_sampling is True, We must specify "nbest_size > 1" or "nbest_size = -1", and "0.0 < alpha < 1.0". "nbest_size = -1" is enabled only on unigram mode and samples from all candidates on the lattice instead of nbest segmentations. 
```

The error message is obviously false (we were able to use `enable_sampling=True` without specifying `nbest_size` and with `alpha==1.0` in the call right above. 

So all in all, how should one do BPE-dropout?

steremma:
Minimally Fixes #569 

Error conditions and messages can be further improved by being model type specific (i.e. alpha has different requirements for LM vs BPE mode).

### DIFF
--- a/python/sentencepiece.i
+++ b/python/sentencepiece.i
@@ -263,7 +263,7 @@ class PySentenceIterator : public sentencepiece::SentenceIterator {
                   nbest_size < 0: assuming that nbest_size is infinite and samples
                     from the all hypothesis (lattice) using
                     forward-filtering-and-backward-sampling algorithm.
-      alpha: Soothing parameter for unigram sampling, and merge probability for
+      alpha: Soothing parameter for unigram sampling, and dropout probability for
         BPE-dropout.
     """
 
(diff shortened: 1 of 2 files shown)


taku910:
Let me close this PR as the fix is already merged.

PR #585, opened 2020-11-19
Remove dependencies on glog

thaink:
This is an effort to make this branch buildable for Android.
- Remove all deps on glog, replacing it by the log implementation on master branch
- Refactor the deps on gflags in src/BUILD

### DIFF
--- a/src/common.h
+++ b/src/common.h
@@ -16,15 +16,111 @@
 #define COMMON_H_
 
 #include <cstring>
+#include <iostream>
 #include <memory>
 #include <string>
 #include <utility>
 #include <vector>
 
-#include <glog/logging.h>
 #include "src/deps/basic_types.h"
 #include "src/deps/canonical_errors.h"
 #include "src/deps/status.h"
 #include "src/deps/status_macros.h"
+
+#if defined(_WIN32) && !defined(__CYGWIN__)
+#define OS_WIN
+#else
+#define OS_UNIX
+#endif
+
+namespace sentencepiece {
+namespace error {
+
+void Abort();
+void Exit(int code);
+void SetTestCounter(int c);
+void ResetTestMode();
+bool GetTestCounter();
+
+class Die {
+ public:
+  explicit Die(bool die) : die_(die) {}
+  ~Die() {
+    std::cerr << std::endl;
+    if (die_) {
+      Abort();
+    }
+  }
+  int operator&(std::ostream &) { return 0; }
+
+ private:
+  bool die_;
+};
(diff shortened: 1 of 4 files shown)


thaink:
@taku910 @resec Could you take a look at this PR?

thaink:
Also update protobuf to v3.9.2 since v3.6.1.2 is deprecated.

thaink:
PTAL @taku910

src/common.h
```diff
@@ -16,15 +16,111 @@
...
+}
+}  // namespace logging
+}  // namespace sentencepiece
+
+#define LOG(severity)                                                     \
+  ::sentencepiece::error::Die(::sentencepiece::logging::LOG_##severity >= \
+                              ::sentencepiece::logging::LOG_FATAL) &      \
+      std::cerr << ::sentencepiece::logging::BaseName(__FILE__) << "("    \
+                << __LINE__ << ") "                                       \
+                << "LOG(" << #severity << ") "
```
broken, line 87: should this order be reversed? I believe bitwise-and reads left to right, so the newline will occur before the message.

src/common.h
```diff
@@ -16,15 +16,111 @@
...
+
+#define CHECK(condition)                                                      \
+  (condition) ? 0                                                             \
+              : ::sentencepiece::error::Die(true) &                           \
+                    std::cerr << ::sentencepiece::logging::BaseName(__FILE__) \
+                              << "(" << __LINE__ << ") [" << #condition       \
+                              << "] "
```
broken, line 94: Here too. Will the program exit before the message is printed?

broken (commented):
thanks for making this update!

src/common.h
thaink: It is fine as Abort only get called in destructor.
The '&' operator ensure the log get printed before that.

src/common.h
thaink: Same as above.

src/common.h
broken: Ah; yes. I completely overlooked that. Thanks.

taku910:
Just a confirmation: Is this PR for bazel branch?  Don't we have any side-effect on master branch?

thaink:
Right. This PR is only for the bazel branch. It will not affect the master branch.

thaink:
Do you think the continuous test fail is related to this PR?

thaink:
Can I have a LGTM?

taku910:
merged. Thank you

PR #623, opened 2021-02-10
add spm proto headers to install

erasaur:
first off, thanks for all your work on this project! 

I came across a small issue trying to run through the example [here](https://github.com/erasaur/sentencepiece/blob/master/doc/api.md#sentencepiecetext-proto) -- currently, if I follow the instructions for installing from source, the protobuf headers aren't installed so the example fails (can't `#include <sentencepiece.pb.h>`). not sure if I'm just missing some steps but I figure I should submit the PR in case this is indeed an oversight.

to be precise, I'm using ubuntu and running these commands for installing (not using builtin proto):
```
% sudo apt-get install cmake build-essential pkg-config libgoogle-perftools-dev libprotobuf-dev protobuf-compiler
% git clone https://github.com/google/sentencepiece.git 
% cd sentencepiece
% mkdir build
% cd build
% cmake -DSPM_USE_BUILTIN_PROTOBUF=OFF ..
% make -j $(nproc)
% sudo make install
% sudo ldconfig -v
```


erasaur:
@googlebot I signed it!

PR #630, opened 2021-02-23
Prevent Zero Width Joiner replaced with whitespace

sarubi:
Resolve google/sentencepiece#629

Currently, the Zero Width Joiner character has been [replaced by whitespace](https://github.com/google/sentencepiece/commit/18c337f32d376b339f15a48c7ccc5c1c288345a1#diff-c0960ff1cd2e917394da837d2a75ca10e06abab61bbd2d36f174fa284c8d700bR57266). Due to that Sinhala language that requires Zero Width Joiner getting altered and resulted in wong output. 

Ideally, we shouldn't replace Zero Width Joiner (200D) with whitespace or empty since it indicates to join two chars without zero width(no whitespace). Also, It requires to present in order to decode the segmentation to raw test successfully. We should keep these special characters as it is.

This PR eliminate the places where WZJ has replaced with whitespace. after referrog to the commit [1],
data/nmt_nfkc.tsv
data/nmt_nfkc_cf.tsv 
builder.cc 

The even-though local build [2] was successful, unfortunately, src/normalization_rule.h didn't get updated, hence I couldn't able to apply my changes. 
Appreciate it if anyone provides insight on this since I wanted to make it work for the Sinhala Language. 

[1] https://github.com/google/sentencepiece/commit/18c337f32d376b339f15a48c7ccc5c1c288345a1#diff-c0960ff1cd2e917394da837d2a75ca10e06abab61bbd2d36f174fa284c8d700bR57266
[2] https://github.com/google/sentencepiece#build-and-install-sentencepiece-command-line-tools-from-c-source

### DIFF
--- a/src/builder.cc
+++ b/src/builder.cc
@@ -368,7 +368,6 @@ util::Status Builder::BuildNmtNFKCMap(CharsMap *chars_map) {
   nfkc_map[{0xFEFF}] = {0x20};  // ZERO WIDTH NO-BREAK
   nfkc_map[{0xFFFD}] = {0x20};  // REPLACEMENT CHARACTER
   nfkc_map[{0x200C}] = {0x20};  // ZERO WIDTH NON-JOINER
-  nfkc_map[{0x200D}] = {0x20};  // ZERO WIDTH JOINER
 
   // Ascii Control characters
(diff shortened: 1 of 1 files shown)


sarubi:
@googlebot I signed it!

Issue #641, opened 2021-03-14
unigram_model_trainer stuck on "Extracting frequent sub strings..."


Talisberg:
Hey,
I'm trying to train the tokenizer and it's stuck on "Extracting frequent sub strings".

I'm using the pip installation for the package, in 2 different machines
1. Windows, i7, 16 cores, 16 RAM
2. linux ubuntu, 16 cores, 120 RAM

args:
```
max_sentence_length: 1000,
vocab_size: 8000,
max_sentencepiece_length: 10,
model_type: UNIGRAM,
num_threads: 16
```

and the same thing has happened in both machines:
1st Test:
a file with 20,000 Sentences.
```
CPU times: user 4.47 s, sys: 84 ms, total: 4.55 s
Wall time: 2.32 s
```

2nd Test
same 20,000 Sentences copied 12 times to the same file
about 240,000 total sentences.

and here it gets stuck on "Extracting frequent sub strings" 
until I just gave up.

so 2 questions:
1. what can I do to improve the running performances
2. under >Fast and lightweight: Segmentation speed is around 50k sentences/sec, and memory footprint is around 6MB.>
what was the environment to get such results?

And Am I missing something?

thiennp223:
For me, just because it takes time to process.

I waited and it runs okay.

AdolfVonKleist:
I have encountered the same issue for some time.  I tracked it down to this line:

- https://github.com/google/sentencepiece/commit/329383b455a5795f3d182159eb0985a3f20f0fa2#diff-51c497698d155a53cbd00dfd1a26018599599affa606d6fd44d663d18d4535aaL116 
   - ` array.push_back(kSentenceBoundary);  // sentence boundary marker.` 

which was removed from:
- https://github.com/google/sentencepiece/blob/7b19d68be0d516d49572a0fa877473df904f90c2/src/unigram_model_trainer.cc#L116

if I understand correctly, the pre-merge version adds an explicit sentence boundary marker for each sentence in the provided training corpus (by default indicated by the newline).  This change effectively treats the entire corpus as a single sentence, and dramatically slows down the training due to the size of the suffix array.  

If you roll back your release to:
 - https://github.com/google/sentencepiece/releases/tag/v0.1.86
 - `pip install sentencepiece==0.1.86` 

[...]

Issue #615, opened 2021-01-12
Missing Unigram 0x0000 delimiter


david-waterworth:
More of a question than a bug, the function `Trainer::MakeSeedSentencePieces()` states

// Merges all sentences into one array with 0x0000 delimiter.

However this isn't performed as far as I can see - the end result seems to be that the input the the ESA algorithm is the concatenation of all sentences (which have been preprocessed to replace ' ' with '▁'

I'm wondering if this is an omission, or was it deemed unnecessary / counter-productive?

The reason I say this is technically I don't believe it's correct concatenate strings using the same end of string token when using Suffix Arrays to enumerate substrings - you're supposed to use a different suffix per string. Otherwise, the ancillary arrays such as the LCP tend to get messed up. For example, if you take "cat" and "bat" and concatenate them as "cat$bat$" you get the sorted suffix array

$
$bat$
at$
at$bat$
bat$
cat$bat$
t$
t$bat$

The longest common prefix of lines 2 and 3 is "at$" - if different terminators were used it would be "at" - what I've observed is this tends to ignore suffixes like "ing" unless they also occur inside words.

david-waterworth:
I found another paper which showed that you can use string concatenation with a suffix array provided the calculation of the LCP (longest common prefix) correctly accounts for the string terminator (i.e. the LCP for "at$" and "at$bat$") should be 2 not 3. But the library used by `sentencepiece` doesn't do this.

It seems to me that by not taking this into account, you're likely to under-represent the frequency of suffixes. For example, if lots of words end in "ed" (followed by  '▁') they're get filtered out when generating the seed vocab for the unigram algorithm. 

It's quite easy to modify the LCP calculation in the esaxx library so that it doesn't count the prefix length beyond the end of a string (denoted either by an inter-word or sentence token). For a technical corpus I've trained a model on it makes a large difference, I suspect for standard languages it's not as important.

taku910:
Fied in v0.1.96. Thank you.

PR #642, opened 2021-03-22
Restore the sentence boundary marker insertion for the unigram model trainer

AdolfVonKleist:
This pull request suggests restoring the sentence boundary insertion code for the suffix array creation in the unigram model trainer, in order to significantly re-speedup training.

This appears to have been deleted here:
- https://github.com/google/sentencepiece/commit/329383b455a5795f3d182159eb0985a3f20f0fa2#diff-51c497698d155a53cbd00dfd1a26018599599affa606d6fd44d663d18d4535aaL116 

during the commit that 'merged the internal sentencepiece'. 

I believe this also relates to the following issues:
- https://github.com/google/sentencepiece/issues/641
- https://github.com/google/sentencepiece/issues/615

I'm not sure if there is a qualitative/modeling reason behind this change (or if there is a settings way to avoid this?), but the difference in training speed with and without this line, for anything larger than very small corpora where the unigram model is concerned is very dramatic.  I'm also not 100% certain that there are not knock-on effects from this change, due to the large number of additional changes in the original commit that removed this line; but so far I do not notice any noticeable difference in downstream accuracy from this restoration.

On the speed point, I downloaded the [librispeech-lm-norm.txt.gz](https://www.openslr.org/resources/11/librispeech-lm-norm.txt.gz) corpus, and extracted the first 500k sentences, after this change, it takes approximately 1min on my MacBook Pro, running in an Ubuntu 20.04 VM via virtualbox to run the following training:
- `spm_train --input=ls-500k.txt --model_prefix=m --vocab_size=2000`

I gave up waiting for it to finish without the above change.

Your thoughts/suggestions on this would be appreciated.

edit: as #615 mentions, the associated comment implies that the boundary is still supposed to be there:
https://github.com/google/sentencepiece/blob/bc53923a9147dc8ffa54034c8ed774de78cc4d39/src/unigram_model_trainer.cc#L108

### DIFF
--- a/src/unigram_model_trainer.cc
+++ b/src/unigram_model_trainer.cc
@@ -119,6 +119,7 @@ TrainerModel::SentencePieces Trainer::MakeSeedSentencePieces() const {
         all_chars[string_util::UnicodeCharToUTF8(c)] += w.second;
       }
     }
+    array.push_back(kSentenceBoundary);  // sentence boundary marker.
   }
 
   CHECK_LE(array.size(),


AdolfVonKleist:
@googlebot I signed it!

eddie-scio:
Could we cut a new release on PyPi?  The latest `0.1.95` does not contain this (very useful) fix, and installing via Github src given the C++ deps isn't very easy.

taku910:
Thank you for your contribution. released 0.1.96 contains this fix.

PR #658, opened 2021-05-25
Fixed mistake in README

George-Ogden:
I've no idea why there was a random s


George-Ogden:
@googlebot I signed it!

PR #666, opened 2021-06-30
Update options.md

felixdae:
fix doc


felixdae:
@googlebot I signed it!

PR #693, opened 2021-09-10
Add FreeBSD support

apraga:
<endian.h> should be <sys/endian.h> instead.


alexDarcy:
@googlebot I signed it!

Issue #700, opened 2021-10-01
Problem compiling in FreeBSD


lpvm:
I'm following the instructions for building the package in a 13.0-RELEASE FreeBSD system.
When  `make -j 10`:

```
[ 11%] Building CXX object src/CMakeFiles/sentencepiece-static.dir/__/third_party/protobuf-lite/coded_stream.cc.o
In file included from /usr/home/myuser/recolhidos/sentencepiece/src/word_model_trainer.cc:20:
/usr/home/myuser/recolhidos/sentencepiece/src/util.h:40:10: fatal error: 'endian.h' file not found
```

`find /usr -name endian.h` return a lot of results.

```
# find /usr/ -name endian.h | grep include
/usr/local/include/winpr2/winpr/endian.h
/usr/local/include/boost/predef/other/endian.h
/usr/src/sys/i386/include/endian.h
/usr/src/sys/amd64/include/endian.h
/usr/src/sys/x86/include/endian.h
/usr/src/sys/arm/include/endian.h
/usr/src/sys/powerpc/include/endian.h
/usr/src/sys/riscv/include/endian.h
/usr/src/sys/mips/include/endian.h
/usr/src/sys/arm64/include/endian.h
/usr/src/contrib/ofed/include/endian.h
/usr/src/tools/build/cross-build/include/linux/endian.h
/usr/src/tools/build/cross-build/include/linux/sys/endian.h
/usr/src/tools/build/cross-build/include/common/machine/endian.h
/usr/src/tools/build/cross-build/include/mac/sys/endian.h
/usr/include/machine/endian.h
/usr/include/infiniband/endian.h
/usr/include/sys/endian.h
/usr/include/x86/endian.h
```
How to deal with this?

xiefangqi:
Hi lpvm, I have a PR to solve this problem, but i have no FreeBSD environment to test, could you help to test this PR https://github.com/google/sentencepiece/pull/710

lpvm:
Hi @xiefangqi, thank you very much for your patch.
It worked fine and I could already install sentencepiece.

As I have other NetBSD and OpenBSD machines, I'll try to compile there and will open an issue if there's some similar (or non-similar) problem.

PR #710, opened 2021-12-06
fix freebsd can't find endian.h problem

xiefangqi:
on FreeBsd, add #include <sys/endian.h>

fix https://github.com/google/sentencepiece/issues/700

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -37,7 +37,11 @@
 #endif
 
 #if !defined(__APPLE__) && !defined(_WIN32)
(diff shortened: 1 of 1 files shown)


xiefangqi:
@taku910 Could you help to merge this commit?

xiefangqi:
@kashif @mattn @kant @danieldk Anyone could help to merge this patch？

taku910:
FreeBSD would be supported with the previous PR. Please let me know if you still see the issue.

PR #715, opened 2021-12-21
fix address sanitizers on clang problem

xiefangqi:
Fix https://github.com/google/sentencepiece/issues/714


xiefangqi:
@taku910 Could you help to merge this commit?

xiefangqi:
@kashif @mattn @kant @danieldk Anyone could help to merge this patch？

PR #719, opened 2022-01-12
Expose CalculateEntropy and SampleEncodeAndScore methods to python

dojoteef:
@kriscao It looks like the code for [You should evaluate your language model on marginal likelihood over tokenisations](https://aclanthology.org/2021.emnlp-main.161/) was added to sentencepiece, but is only exposed through the C++ API. This pull request exposes the `CalculateEntropy` and `SampleEncodeAndScore` methods to python to make it easier to use.

This change necessitated running SWIG, which also updated __init__.py and the .cxx wrapper automatically.

I've tested out these changes locally in the python interpreter where they appear to behave as expected.

NOTE: I have already signed the Google Individual Contributor License Agreement.

### DIFF
--- a/python/src/sentencepiece/__init__.py
+++ b/python/src/sentencepiece/__init__.py
@@ -113,6 +113,12 @@ class SentencePieceProcessor(object):
     def SampleEncodeAsIds(self, input, nbest_size, alpha):
         return _sentencepiece.SentencePieceProcessor_SampleEncodeAsIds(self, input, nbest_size, alpha)
 
+    def SampleEncodeAndScoreAsPieces(self, input, samples, theta, wor, include_best):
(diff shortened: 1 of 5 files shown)


kriscao:
Hi @dojoteef, thanks for doing this! It's probably best to wait for @taku910 (the repository owner) to take a look at this PR, and give his opinion.

taku910:
Thank you.

Given this PR, I manually added the support of Entropy, SampleAndScore and NBestEncode methods.
In addiont,  add_bos/eos/reverse logic are moved from python to C++ so we can improve the peformance.

https://github.com/google/sentencepiece/blob/master/python/src/sentencepiece/sentencepiece.i#L466

PR #820, opened 2023-02-17
Fix setup-python version not detected

juliusfrost:
Fixes this error:
Build wheels on ubuntu-latest
Unable to resolve action `actions/setup-python@v4.5`, unable to find version `v4.5`

https://github.com/google/sentencepiece/actions/runs/4204923860


juliusfrost:
@taku910 This should fix the failing build on the master branch. Thanks

PR #823, opened 2023-02-21
Add interface for C# calling

zhongkaifu:
@zamgi

As I remembered, you added this code to a separated SentencePiece package that C# code, such as Seq2SeqSharp, is able to call SentencePiece. Can you please create a pull request to integrate it into official SentencePiece project? Or would you mind if I do it ? Thanks.


zamgi:
> 
Hi @zhongkaifu 

"Or will you mind if I do?"
- yes please do it.

zamgi:
> 
@zhongkaifu 

maybe add the C-sharp wrapper code there right away? (or not?)

zhongkaifu:
@taku910  Could you please take a look this PR?

We have some C# code depends on SentencePiece. Currently, we modified SentencePiece by adding code in this PR, so that C# code could P/Invoke to call functions in SentencePiece. But that means when we ship our product, we have to ship the modified SentencePiece as well. It would be great if these C# calling interfaces could be integrated into the offical SentencePiece, so everyone could use it and we can also let our product to depends on the official SentencePiece only.

Thanks
Zhongkai Fu


Thanks
Zhongkai Fu

taku910:
Seems that you want to add these pure C functions to build Windows DLL.  We can see these intentions, but  we won't need to include them in the main sentencepiece_processor.{h.cc}. These added functions are just calling C++ public functions, so we can simply make another pure .h and *.c functions and include them in the build targets.

zhongkaifu:
Hi @taku910 

Thanks for your suggestion.
This code is able to run at any platform that .NET supports, such as Windows, Linux, MacOS and others. I already tested it on different platforms (Windows/Linux) and architectures (x86/x64/ARM64)

As your suggestion, I created two new separated files only for .NET interfaces implementation and added them into the build targets file. Can you please take a look? Thanks in advance.

Thanks
Zhongkai Fu

taku910:
I'm afraid that we still don't see why these interfaces must be included in the sentencepiece. They are semantically independent since they all just call the public APIs.  It would be possible to release an independent C# wrapper library as long as it is built upon the public APIs.

https://github.com/wang1ang/SentencePieceWrapper

zhongkaifu:
Fair enough. Closed it. Thanks.

PR #920, opened 2023-10-19
Better External Abseil and Protobuf Linkage Support

Halmoni100:
Code changes to assist with integrating sentencepiece in cpp projects as a shared library linked to google open source dependencies externally built as shared libraries.

I tried to follow what seemed like google open source cmake conventions for the cmake options.

Code changes include:
- Being more careful with isolating namespaces, include paths, and potential global namespace conflicts
- Using externtal absl headers in place of internal third_party headers
- Change minloglevel ABSL_FLAG to spm_minloglevel to avoid global namespace minloglevel in grpc


Halmoni100:
Also I kept the internal implementation of absl random because that seemed significant, I'm not really sure what the plan is there

PR #938, opened 2023-11-17
Set minimal permissions for GitHub workflows

### LINKED ISSUE #937: Ensure workflows run with minimal permissions
Hi, it's Pedro and I'm back (see https://github.com/google/sentencepiece/pull/934) with another security suggestion!

**Describe the bug**
SentencePiece's automated workflows currently run with write-all tokens. This puts the project at risk of supply-chain attacks. [GitHub](https://docs.github.com/en/actions/security-guides/security-hardening-for-github-actions#using-secrets) recommends ensuring all workflows run with minimal permissions.

I've taken a look at the workflows and they don't need broad permissions.

This issue can be solved in two ways:
- add top-level read-only permissions to all workflows; and/or
- set the default token permissions to read-only in the repo settings.

I'll be sending a PR along with this issue that sets the top-level permissions (and grants additional permissions for the jobs that need them). If you instead (or also) wish to modify the default token permissions:

1. Open the repo settings
2. Go to [Actions > General](https://github.com/google/sentencepiece/settings/actions)
3. Under "Workflow permissions", set them to "Read repository contents and packages permissions"

pnacht:
Fixes #937.

This PR sets read-only top-level permissions on all SentencePiece workflows. The jobs that add artifacts to releases (in `cmake.yml` and `wheel.yml`) are granted additional permissions at the job level for future-proofing in case new jobs are ever added to the workflow.

### DIFF
--- a/.github/workflows/cifuzz.yml
+++ b/.github/workflows/cifuzz.yml
@@ -1,5 +1,9 @@
 name: CIFuzz
 on: [pull_request]
+
+permissions:
+  contents: read
+
 jobs:
   Fuzzing:
     runs-on: ubuntu-latest
--- a/.github/workflows/cmake.yml
+++ b/.github/workflows/cmake.yml
@@ -8,6 +8,9 @@ on:
   pull_request:
     branches: [ master ]
 
+permissions:
+  contents: read
+
 jobs:
   build:
     strategy:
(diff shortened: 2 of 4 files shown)


pnacht:
It's not quite on-topic for this PR, but I've bumped `cibuildwheel` to the latest version ([2.16.2](https://github.com/pypa/cibuildwheel/releases)) to fix the failing `wheels.yml` for MacOS.

I noticed the dependabot PR #936 is also failing that test for the same reason. If this PR is merged, I suggest commenting "@dependabot rebase" on that PR to fix the failing test there as well.

Issue #876, opened 2023-06-06
Patches carried by conda-forge for packaging sentencepiece


h-vetinari:
Hi there - I help package many a package for conda-forge, which is a pretty big ecosystem of packages especially in data science and ML/AI. It started out mainly focussed on Python, but has come to include many other languages as well.

One strong point (especially in comparison to pip) is that we are able to sanely redistribute non-Python artefacts. We can track the relevant ABI of a package and rebuild its descendant packages where necessary. This allows us to massively to (almost) exclusively distribute shared libraries, which lowers package footprints, and has several other benefits.

We've been shipping sentencepiece for about 3 years through the so-called [feedstock](https://github.com/conda-forge/sentencepiece-feedstock), and packages can be installed like:
```
conda install -c conda-forge sentencepiece  # or mamba instead of conda
```

In the context of wanting to [package](https://github.com/conda-forge/staged-recipes/pull/17129) [torchtext](https://github.com/pytorch/text) (which depends on sentencepiece -- but really only the C++ bits) I wanted to split off a separate "output" / "package" called `libsentencepiece` that torchtext can build against, without having to spuriously rebuild it per python version (it would literally be the same in each case).

Also, crucially, we absolutely need to rely on our own abseil & protobuf builds, which became a pretty big sticking point. I finally managed to pull off this [split](https://github.com/conda-forge/sentencepiece-feedstock/pull/33) last year (there's references there to my very long-winded previous tries; eventually I gave up on trying to build a shared library on windows because it was getting too insane[^1]).

[^1]: by that point I was already patching the output generated by protobuf for some silly constexpr errors

However, at the time, I ran out of steam to raise an issue here, and discuss how - if at all - it would be possible to upstream some of these changes. That's all the more the case because some of the patches I came up with are definitely not ready for primetime here. I got quite exasperated with how the code in [`thirdparty/absl`](https://github.com/google/sentencepiece/tree/master/third_party/absl) was neither third-party nor abseil, so I ended up replacing that wholesale. Of course that doesn't mean I'm suggesting to do this here verbatim.

However, the abseil situation is perhaps one of the biggest sticking points. There's a variable called `SPM_USE_EXTERNAL_ABSL`, but it really _doesn't_ do what it says on the tin, because all the include paths still point to `third_party/absl`, rather than abseil proper. This issue has also been encountered by other people: #869.

[...]

PR #979, opened 2024-02-20
move setting of default CMAKE_INSTALL_{BIN,INCLUDE,LIB}DIR before first use

h-vetinari:
One of the things I had misdiagnosed in #876 was how I was getting `lib64` and other such things into the build, when nothing should have been setting them like that.

Now after rebasing the patches to 0.2.0, I realize that this is simply due to the fact that the default values (in case the various `CMAKE_INSTALL_*` quantities are not explicitly overriden) are defined _after_ their first use further up in `CMakeLists.txt`:
https://github.com/google/sentencepiece/blob/17d7580d6407802f85855d2cc9190634e2c95624/CMakeLists.txt#L60-L66

Fix this by moving the setting of the default values up before they're first used.

Also rename `CMAKE_INSTALL_INCDIR` -> `CMAKE_INSTALL_INCLUDEDIR`, which corresponds to the GNUInstallDirs default that CMake itself [documents](https://cmake.org/cmake/help/latest/command/install.html).


fundawang:
There is another problem here.  The default installation directories under linux are defined in `GNUInstallDirs`. So `include(GNUInstallDirs)` should be referenced before it has been used first time. i.e.:
```
if(UNIX)
  include(GNUInstallDirs)
endif()

if (NOT DEFINED CMAKE_INSTALL_BINDIR)
  set(CMAKE_INSTALL_BINDIR bin)
endif()

if (NOT DEFINED CMAKE_INSTALL_LIBDIR)
  set(CMAKE_INSTALL_LIBDIR lib)
endif()

...

if (UNIX)
  set(prefix ${CMAKE_INSTALL_PREFIX})
  set(exec_prefix "\${prefix}")
  set(libdir "\${exec_prefix}/${CMAKE_INSTALL_LIBDIR}")
  set(includedir "\${prefix}/${CMAKE_INSTALL_INCLUDEDIR}")
else()
  set(prefix ${CMAKE_INSTALL_PREFIX})
  set(exec_prefix "\${prefix}")
  set(libdir "\${exec_prefix}/lib")
  set(includedir "\${prefix}/include")
endif()
```

PR #1013, opened 2024-05-22
bump CMake minimum required version to avoid warnings

alexlnkp:
fixes future deprecation
```
CMake Deprecation Warning at lib/sentencepiece/CMakeLists.txt:15 (cmake_minimum_required):
  Compatibility with CMake < 3.5 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value or use a ...<max> suffix to tell
  CMake that the project does not need compatibility with older versions.
```

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.!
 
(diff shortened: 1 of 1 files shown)


alexlnkp:
I'm using a github provided EMail address for commits, what do you want me to do?

PR #1022, opened 2024-06-05
Parameterize lattice node allocator size to optimize chunk allocation performance

PriyankaRanganath:
Lattice node_allocator_ allocates chunks of 1024 items, which is 49 152 bytes total.
This huge size of chunk causes performance issues in high qps environments.
Our internal tests shows significant latency reduction is we can parametrize default chunk size for your needs.

### DIFF
--- a/src/bpe_model.cc
+++ b/src/bpe_model.cc
@@ -36,7 +36,7 @@ Model::Model(const ModelProto &model_proto) {
 Model::~Model() {}
 
 std::vector<std::pair<absl::string_view, int>> Model::SampleEncode(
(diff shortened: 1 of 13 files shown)


PriyankaRanganath:
hi @taku910 ,

Is it possible to review this PR?
Thanks,
P

taku910:
We are afraid that it cannot be merged because it is only effective in special sampling situations.  The lattice is not used in normal encode function.  We will not be able to manage this new feature.

We cannot guarantee that this function is merged, but it would be appreciated if you first create an issue to broadly discuss  the performance, implementation, and the importance of this feature.

PR #1106, opened 2025-04-10
Migrate third_party/ to git submodules.

ZacharyGarrett:
This replaces `third_party/absl/` and `third_party/protobuf-lite/` with git submodules checked out at specific versions for `abseil-cpp`, `googletest`, `benchmark`, and `protobuf` libraries.

This change allows removing local code copies of absl and googltest functionality.


taku910:
Thank you for your contribution.

We cannot accept such a large-scale change at this time. It would require a significant amount of time to verify and pose a high risk. Please focus on making smaller, more targeted changes.

Issue #1110, opened 2025-04-21
Support Python 3.13t free-threading wheel distributions


andfoy:
As some of you might be aware, since CPython 3.13, there is experimental support for an alternative interpreter distribution (via [PEP703](https://peps.python.org/pep-0703/)) that disables the GIL altogether, thus enabling support for real thread parallelism.

As of now, sentencepiece can be build without any issues on 3.13t, however, more testing is required in order to ensure free-threading compatibility, which can be summarized in performing the following tasks:

- [ ] Audit Python bindings and declare them free-threading compatible (xref https://py-free-threading.github.io/porting/#updating-extension-modules).
- [ ] Run the test suite with pytest-run-parallel to find potential issues, and fix them.
- [ ] Run the test suite under ThreadSanitizer. Should be feasible, since there are no runtime dependencies aside from Cython.
- [ ] Add cp313t-* to CI to build free-threading wheels. 

I can volunteer in taking charge of these tasks if necessary

crusaderky:
This stack of PRs introduces free-threading support:

- #1120 
- #1123
- #1125
- #1126

I have not run ThreadSanitizer but I think it's overkill. pytest-run-parallel already tests the one use case that matters, which is when the same object is used to process multiple strings from different threads in parallel.

~I'm now investigating an issue with SWIG that prevents me from regenerating the binding with the `-threads` flag (see caveat in #1126).~ [EDIT] fixed

crusaderky:
https://github.com/swig/swig/issues/3212

crusaderky:
All outstanding caveats fixed; this issue is ready to be closed as soon as the PRs above are merged.

taku910:
Fixed in https://github.com/google/sentencepiece/releases/tag/v0.2.1

reneleonhardt:
Thank you Guido for all your hard work to bring a piece of the AI puzzle into the 21st century 🚀

PR #1112, opened 2025-04-30
Add support for free-threaded CPython

andfoy:
See #1110

This PR enables support for free-threaded CPython 3.13, most of the changes are related to testing, which enable parallel testing using pytest-run-parallel. Additionally, ThreadSanitizer checks were performed without any issue

### DIFF
--- a/python/test/sentencepiece_test.py
+++ b/python/test/sentencepiece_test.py
@@ -24,11 +24,15 @@ import io
 import os
 import pickle
 import unittest
+import threading
 import sentencepiece as spm
+
+HERE = os.path.dirname(os.path.abspath(__file__))
+print(HERE)
 
 print('VERSION={}'.format(spm.__version__))
 
-data_dir = 'test'
+data_dir = HERE
 if sys.platform == 'win32':
   data_dir = os.path.join('..', 'data')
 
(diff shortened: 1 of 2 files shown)


python/test/sentencepiece_test.py
```diff
@@ -24,11 +24,15 @@
 import os
 import pickle
 import unittest
+import threading
 import sentencepiece as spm
 
+HERE = os.path.dirname(os.path.abspath(__file__))
+print(HERE)
```
rgommers, line 31: debug print left in

python/test/sentencepiece_test.py
```diff
@@ -24,11 +24,15 @@
 import os
 import pickle
 import unittest
+import threading
 import sentencepiece as spm
 
+HERE = os.path.dirname(os.path.abspath(__file__))
```
rgommers, line 30: This change looks right, to make it possible to run the tests from any directory rather than only in `python/`.

There are path-related changes that are not right though - after a test run with this branch, many `m_xxx.model` and m_xxx.vocab` files are left behind in `python/`. It'd be good to clean those up - or otherwise at a minimum extend the pattern for these files in the top-level `.gitignore`.

python/test/sentencepiece_test.py
```diff
@@ -232,32 +238,32 @@ def test_train_iterator(self):
     # suppress logging (redirect to /dev/null)
     spm.SentencePieceTrainer.train(
         input=os.path.join(data_dir, 'botchan.txt'),
-        model_prefix='m',
+        model_prefix=f'm_{tid}',
         vocab_size=1000,
-        logstream=open(os.devnull, 'w'),
+        # logstream=open(os.devnull, 'w'),
```
rgommers, line 243: This doesn't look right.

rgommers (changes requested):
This is a pretty WIP PR at the moment it looks like:

- The test changes leave files behind (quite a lot when testing with `pytest-run-parallel`).
- The extension module isn't marked as safe to run without the GIL. To achieve this, it looks like you need to regenerate `sentencepiece_wrap.cxx` with a version of SWIG that includes https://github.com/swig/swig/pull/3137. 
- Is there a log for a successful CI run?
- The CLA needs signing before this PR can move forward.

crusaderky:
Superseded by https://github.com/google/sentencepiece/issues/1110#issuecomment-3024357525

crusaderky:
@taku910 this PR was not fit for purpose. Please revert.

Issue #1111, opened 2025-04-22
SentencePiece Installation: Build Wheel Failed to Run Successfully


Kax-Industries:
I've been attempting to install SentencePiece for some time now, but I have been unsuccessful so far. I've tried several troubleshooting steps, including redownloading CMake and using the -DCMAKE_POLICY_VERSION_MINIMUM=3.5, but the issue persists. I would appreciate any assistance in resolving this problem. Here is the error code I'm encountering: 

Collecting sentencepiece
  Using cached sentencepiece-0.2.0.tar.gz (2.6 MB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
  
  × Getting requirements to build wheel did not run successfully.
  │ exit code: 1
  ╰─> [37 lines of output]
      -- Building for: NMake Makefiles
      CMake Error at CMakeLists.txt:15 (cmake_minimum_required):
        Compatibility with CMake < 3.5 has been removed from CMake.
      
        Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
        to tell CMake that the project requires at least <min> but has been updated
        to work with policies introduced by <max> or earlier.

        Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.

      -- Configuring incomplete, errors occurred!
      Traceback (most recent call last):
        File "C:\Users\Kax\AppData\Local\Programs\Python\Python313\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 389, in <module>
          main()
          ~~~~^^
        File "C:\Users\Kax\AppData\Local\Programs\Python\Python313\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 373, in main
          json_out["return_val"] = hook(**hook_input["kwargs"])
                                   ~~~~^^^^^^^^^^^^^^^^^^^^^^^^
        File "C:\Users\Kax\AppData\Local\Programs\Python\Python313\Lib\site-packages\pip\_vendor\pyproject_hooks\_in_process\_in_process.py", line 143, in get_requires_for_build_wheel
          return hook(config_settings)
        File "C:\Users\Kax\AppData\Local\Temp\pip-build-env-n8xytm7x\overlay\Lib\site-packages\setuptools\build_meta.py", line 331, in get_requires_for_build_wheel
          return self._get_build_requires(config_settings, requirements=[])
[...]

PR #1115, opened 2025-05-09
setup.py fix

teamclouday:
Trying to fix setup code for two errors when installing the package:

Error 1: cmake_minimum_required compatibility
```
CMake Error at CMakeLists.txt:15 (cmake_minimum_required):
Compatibility with CMake < 3.5 has been removed from CMake.

Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
to tell CMake that the project requires at least <min> but has been updated
to work with policies introduced by <max> or earlier.

Or, add -DCMAKE_POLICY_VERSION_MINIMUM=3.5 to try configuring anyway.
```

Error 2: source directory not found
```
CMake Error: The source directory "xxx" does not exist.
```

Only tested on Windows python 3.13. Related issue: #1111

### DIFF
--- a/python/setup.py
+++ b/python/setup.py
@@ -144,7 +144,7 @@ if os.name == 'nt':
       cmake_arch = "ARM64"
     subprocess.check_call([
         'cmake',
-        'sentencepiece',
+        '..',
         '-A',
         cmake_arch,
         '-B',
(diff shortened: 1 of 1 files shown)


Ahajha:
This is already fixed @HEAD - the CMake minimum version is set to 3.5. The issue is that they haven't released a new version since that.

teamclouday:
I see the PR now #1013 Hope they can make a new release. Many packages depend on this

Issue #1108, opened 2025-04-21
Cannot build with GCC 15


megascrapper:
I was trying to build `sentencepiece` with GCC 15 but got build errors relating to missing `uint32_t` type. It builds just fine with GCC 14 so there's probably some changes between GCC 14 and 15 relating to auto-importing `cstdint`.

<details>

<summary>Build output</summary>

```plaintext
$ cmake ..
CMake Deprecation Warning at CMakeLists.txt:15 (cmake_minimum_required):
  Compatibility with CMake < 3.10 will be removed from a future version of
  CMake.

  Update the VERSION argument <min> value.  Or, use the <min>...<max> syntax
  to tell CMake that the project requires at least <min> but has been updated
  to work with policies introduced by <max> or earlier.


-- VERSION: 0.2.1
-- The C compiler identification is GNU 15.0.1
-- The CXX compiler identification is GNU 15.0.1
-- Detecting C compiler ABI info
-- Detecting C compiler ABI info - done
-- Check for working C compiler: /usr/sbin/cc - skipped
-- Detecting C compile features
-- Detecting C compile features - done
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /usr/sbin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Success
-- Found Threads: TRUE
-- Found TCMalloc: /usr/lib64/libtcmalloc_minimal.so
-- Configuring done (0.4s)
-- Generating done (0.0s)
-- Build files have been written to: /tmp/sentencepiece/build

$ make -j $(nproc)
[  4%] Building CXX object src/CMakeFiles/sentencepiece_train-static.dir/trainer_interface.cc.o
[  4%] Building CXX object src/CMakeFiles/sentencepiece_train-static.dir/unigram_model_trainer.cc.o
[  4%] Building CXX object src/CMakeFiles/sentencepiece_train-static.dir/word_model_trainer.cc.o
[  8%] Building CXX object src/CMakeFiles/sentencepiece_train-static.dir/char_model_trainer.cc.o
[  4%] Building CXX object src/CMakeFiles/sentencepiece-static.dir/__/third_party/protobuf-lite/arena.cc.o
[...]

Issue #1113, opened 2025-05-07
pipx install sentencepiece fails to install


gonsolo:
1. Fix with gcc 15 (include cstdint)
2. Release a new version and upload to Pypi, so that CMake 4 works

gonsolo:
```
diff --git a/src/sentencepiece_processor.h b/src/sentencepiece_processor.h
index dd3f092..010d500 100644
--- a/src/sentencepiece_processor.h
+++ b/src/sentencepiece_processor.h
@@ -15,6 +15,7 @@
 #ifndef SENTENCEPIECE_PROCESSOR_H_
 #define SENTENCEPIECE_PROCESSOR_H_
 
+#include <cstdint>
 #include <cstring>
 #include <memory>
 #include <string>

```

Ahajha:
Thanks for reporting this here too. This definitely is going to be a pain point going forward as more systems update CMake and GCC versions.

polarathene:
Seems there's been a PR with that fix for a few weeks now (April 22): https://github.com/google/sentencepiece/pull/1109

The author hasn't signed the CLA apparently which might be why no review has taken place 🤷‍♂ 

No new releases since Feb 2024 though? Ideally a new release is tagged after that's resolved.

taku910:
This issue is fixed on the head master. Thank you.

gonsolo:
> This issue is fixed on the head master. Thank you.

No it's not. You fixed 1. (GCC issue), but 2. (Release and upload to Pypi) is not done. 👎 
Until that is done, all packages relying on sentencepiece (Modular, mojo) are broken.
Unfortunately I'm not allowed to reopen my own issue. 👎

Ahajha:
For modular/mojo/max, the dependency was removed, at least in the last few weeks of nightlies, and the next stable will also have the dependency removed.

gonsolo:
> For modular/mojo/max, the dependency was removed, at least in the last few weeks of nightlies, and the next stable will also have the dependency removed.

Ok. But that doesn't change the fact that the fixed cstdint is worthless without a release.

Ahajha:
Yes, agreed.

PR #1119, opened 2025-06-02
Add missing <cstdint> in sentencepiece_processor.h

jmsunseri:
Fixes "uint32_t does not name a type" gemma.cpp build errors....


jmsunseri:
this might close issue #1108 and #1113?

PR #1122, opened 2025-06-17
I've added support for Python 3.13.

AureliusIvan:
This change adds support for Python 3.13 by:

- Updating cibuildwheel and its dependencies to the latest versions.
- Adding the Python 3.13 classifier to setup.py.


crusaderky:
Please use #1120 instead.

taku910:
Thanks for the change. We think this change is already included in the latest head.

PR #1123, opened 2025-06-26
Use modern Pytest syntax

crusaderky:
This PR builds on top of #1120 and is propaedeutic to #1110. 
Please see only the last commit.

- One-off run of `ruff format`
- Ditch the legacy unittest syntax for a much more modern pytest syntax.
- Don't write temporary files to project directory.


taku910 (approved):

crusaderky:
@taku910 this PR depends on #1120; could you start from that one?

PR #1124, opened 2025-06-28
Add support for Python 3.13

celsowm:
This commit updates the setup.py classifiers to include Python 3.13. No changes are needed for the GitHub Actions workflow as cibuildwheel will automatically pick up Python 3.13 when it becomes available in the runners.


crusaderky:
Please use #1120 instead.

taku910 (approved):

PR #1125, opened 2025-06-30
Regen SWIG/protobuf; Refactor Python utils; clean up exported symbols

crusaderky:
- Towards #1110
- Update to latest released version of protobuf for the sake of code sanity
- Update to SWIG git tip @ 2025-07-13 (4.4.0devXXX). This is a prerequisite for #1110. Note that the latest released version, 4.3.1, segfaults.
- Add development scripts to regenerate SWIG and protobuf files again in the future
- Refactor a bunch of internal utilities out of the SWIG template and to a proper Python file
- `from sentencepiece import *` will no longer import a wealth of internal symbols.

CI needs #1120 to complete. This is otherwise an independent PR.
CI run: https://github.com/crusaderky/sentencepiece/pull/4


python/src/sentencepiece/__init__.py
```diff
@@ -1,1224 +1,29 @@
...
-
+from .sentencepiece import (
+    ImmutableNBestSentencePieceText,
+    ImmutableSentencePieceText,
+    ImmutableSentencePieceText_ImmutableSentencePiece,
+    SentencePieceNormalizer,
+    SentencePieceProcessor,
+    SentencePieceTrainer,
+    SetMinLogLevel,
+    SetRandomGeneratorSeed,
```
crusaderky, line 9: Not sure if all of these are meant to be public API. @taku910 should know better.

PR #1135, opened 2025-08-08
optimization/simplification: Unigram pruning does not need inverted array

sanderland:
The prior `inverted` structure was not a set (it counted all Viterbi occurrences per sentence), despite the comment.

Taking that as intentional:
* The inverted array is actually not needed, as F=freq[i]/vsum
* The normalization should probably just use 'sum' rather than 'vsum', although either factor has no effect on the order of the loss array or results.

Tests: 
* C++ tests passing locally; example training on data/botchan.txt succeeds.
* Python tests: pass

### DIFF
--- a/src/unigram_model_trainer.cc
+++ b/src/unigram_model_trainer.cc
@@ -431,23 +431,16 @@ TrainerModel::SentencePieces Trainer::PruneSentencePieces(
     }
   }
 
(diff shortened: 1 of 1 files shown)


sanderland:
@taku910 any thoughts on this suggested change? Thanks :)

PR #1140, opened 2025-08-20
Update util.cc

Coder-Harshit:
int -> std::size_t

=> warning FIX related to type and int comparison


Coder-Harshit:
Signed the CLA

PR #1157, opened 2025-10-21
1156 Update license in setup.cfg

### LINKED ISSUE #1156: Python package does not have license
**Describe the bug**
Package does not have license in the metadata.
https://pypi.org/project/sentencepiece/ does not show the license

**To Reproduce**
Open https://pypi.org/project/sentencepiece/ 

or run sentencepiece

Image

wagenrace:
### Pull Request Details

This will add the Apache-2.0 according to the SPDX standard and fixes #1156

### DIFF
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -10,6 +10,7 @@ authors = [
 ]
 description = "Unsupervised text tokenizer and detokenizer."
 readme = "README.md"
+license = { text = "Apache-2.0" }
 requires-python = ">=3.9"
 urls = { "Homepage" = "https://github.com/google/sentencepiece" }
 classifiers = [
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -10,7 +10,8 @@ authors = [
 ]
 description = "Unsupervised text tokenizer and detokenizer."
 readme = "README.md"
-license = { text = "Apache-2.0" }
+license = "Apache-2.0"
+license-files = ["../LICENSE"]
 requires-python = ">=3.9"
 urls = { "Homepage" = "https://github.com/google/sentencepiece" }
 classifiers = [
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -1,3 +1,3 @@
 [metadata]
 description_file = README.md
-license = Apache-2.0
+
--- a/python/setup.cfg
+++ b/python/setup.cfg
@@ -1,2 +1,3 @@
 [metadata]
 description_file = README.md
+license = Apache-2.0


tkng:
I noticed the same issue and came here. I'm glad to see that this pull request has already proposed a fix — thank you for working on this.

It seems that the addition of license information in this pull request follows an deprecated approach, according to the [Python Packaging User Guide](https://packaging.python.org/en/latest/guides/writing-pyproject-toml/#license-and-license-files).

It would be better to update it in accordance with [PEP 639](https://peps.python.org/pep-0639/), like this:

```diff
diff --git a/python/pyproject.toml b/python/pyproject.toml
index 4ec78b6..a646638 100644
--- a/python/pyproject.toml
+++ b/python/pyproject.toml
@@ -11,6 +11,8 @@ authors = [
 description = "Unsupervised text tokenizer and detokenizer."
 readme = "README.md"
 requires-python = ">=3.9"
+license = "Apache-2.0"
+license-files = ["../LICENSE"]
 urls = { "Homepage" = "https://github.com/google/sentencepiece" }
 classifiers = [
   "Programming Language :: Python :: 3",
```

What do you think about this change?

wagenrace:
@tkng that also works. I think that is an even clear methods but have to read up on the path of License file
You now have the path the License file within the repo
But might be you need to path the license within the module

tkng:
@wagenrace I built a wheel file after applying the patch above, extracted its contents, and confirmed that the license file exists at `sentencepiece-0.2.2.dist-info/licenses/LICENSE`. Hence, we can use the notation `license-files = ["../LICENSE"]`.

wagenrace:
@tkng  Great, I updated it
Good to know for myself too :smile_cat:

tkng:
Sorry, when I tried building with `python -m build` using the latest setuptools, I got the following warning:

```
/tmp/build-env-ki0ozy_3/lib/python3.12/site-packages/setuptools/dist.py:483: SetuptoolsDeprecationWarning: Pattern '../LICENSE' cannot contain '..'
!!

        ********************************************************************************
        Please ensure the files specified are contained by the root
        of the Python package (normally marked by `pyproject.toml`).

        By 2026-Mar-20, you need to update your project and remove deprecated calls
        or your builds will no longer be supported.

        See https://packaging.python.org/en/latest/specifications/glob-patterns/ for details.
        ********************************************************************************
```

Currently, the notation “../LICENSE” is allowed, but will not be allowed in the future... :cry:  It seems better to have a symbolic link to the top level `LICENSE` file, as this made the warning disappear.

wagenrace:
@tkng ironically, a deprecation warning of SetupTools is what brought me here in the first place.  
I had to update my tool license_scanner to remove SetupTools and was testing it on the 1000 most downloaded Python packages. That is how I found this typo in SentencePiece 😆

wagenrace:
https://github.com/pypa/setuptools/issues/2339
It seems like setuptools is not really supporting file from the top level.
I see 2 options:
- Copy the LICENSE file into the Python folder
- Set the build location to the root of the repo and change the manifist.in to use only the Python except for the license

Personally, I think option 1 is clearer

tkng:
I found that the issue with `license-files = ["../LICENSE"]` can be solved by using a symbolic link. However, we have another problem, the CI is failing now. We need to determine why and fix it.

My investigation shows that the failure comes from the `packaging` package being stuck at version 24.1, which doesn't support the new license expression format. (Twine relies on `packaging` when validating wheels.) Because of this, we need to upgrade `packaging` to version 25.0 or later.

There seems two reasonable ways to proceed:

1. Add `packaging>=25.0` as a dependency in `base.in`.
2. Run pip-compile with `--upgrade` option to refresh `base.txt` and `cibuildwheel.txt`. Without `--upgrade`, pip-compile tries to stay close to what’s already in `base.txt`, so the `packaging` version never moves up.

I lean toward the second option, though I don't have strong opinion. I think it would be best to let the author, @taku910, decide. What do you think?

wagenrace:
Fixed in other pr

PR #1167, opened 2025-11-25
Avoid directly calling setup.py, migrate to pypa/build

### LINKED ISSUE #1166: build_sdist.sh calling `setup.py` directly
According to this blog post, https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html, we should not call setup.py directly. However, in `python/build_sdist.py`, `setup.py` is called directly now.

Actually, sometimes we can see the warning message like below when we're calling setup.py directly:

```
********************************************************************************
   Please avoid running ``setup.py`` directly.
   Instead, use pypa/build, pypa/installer or other
   standards-based tools.

   By 2025-Oct-31, you need to update your project and remove deprecated calls
   or your builds will no longer be supported.

   See https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html for details.
********************************************************************************
```

At this point the issue has no practical impact, but I believe it is something that should be addressed sooner rather than later.

tkng:
Direct invocations of setup.py have been deprecated by the Python packaging community and will be unsupported after 2025-Oct-31. This commit migrates the build process to use `python -m build`, which is the new recommended approach.

### Changes:
- Add `build` package to workflow requirements
- Replace `build_sdist.sh` script with `python -m build --sdist`
- Update README installation instructions to use `python -m build`

### References:
- https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html
- https://setuptools.pypa.io/en/latest/deprecated/commands.html

### Testing Information:
As described in #1166, this problem doesn't have any practical impact yet, hence there's no procedure to cause an actual problem.

Fix #1166.

### DIFF
--- a/.github/workflows/wheel.yml
+++ b/.github/workflows/wheel.yml
@@ -82,7 +82,7 @@ jobs:
       - name: Build sdist archive
         working-directory: python
         run: |
-          python -m build --sdist
+          sh build_sdist.sh
 
       - name: Fetch sdist archive
         uses: tj-actions/glob@2deae40528141fc53131606d56b4e4ce2a486b29 # v22.0.2
--- a/python/build_sdist.sh
+++ b/python/build_sdist.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+mkdir -p sentencepiece
+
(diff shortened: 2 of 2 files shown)


tkng:
@taku910 I've fixed the build error on CI (though I can't be 100% certain since I don't have a Windows on Arm machine.) Could you please approve the pending CI again?

PR #1174, opened 2025-12-10
Add Dockerfile containing the spm command line tools

kkew3:
This PR adds a Dockerfile that builds an image containing the `spm_*` C++ command line tools, which could make deployment and the use of these command line tools easier than building and installing locally.

The image is based on [`alpine:3.23`](https://hub.docker.com/_/alpine). It builds in two stages to ensure small final image size.

The image has been built and run on my side without error.

# Why this PR

On dockerhub there has been a variety of third-party sentencepiece images like [smizy/sentencepiece](https://hub.docker.com/r/smizy/sentencepiece). However, these are mostly outdated versions of sentencepiece. Therefore, building docker image from the latest sentencepiece source should be made more readily available.

# Usage

After running

```bash
docker build -t sentencepiece .
```

locally, the command line tools can be run by e.g.

```bash
docker run --rm sentencepiece spm_train --help
```

### DIFF
--- a/contrib/docker/README.md
+++ b/contrib/docker/README.md
@@ -1,0 +1,20 @@
+# Dockerfile
+
+This directory contains a convenient Dockerfile to build and run sentencepiece C++ command line tools in a container.
+**IMPORTANT**: this Dockerfile is community-maintained on a *best effort* basis only.
+The sentencepiece core team does not take responsibility for ongoing maintenance, security fixes, or guarantees of uptime/compatibility.
+Use at your own risk.
+
+## Quickstart
+
+Build locally (under the repository root directory):
+
+```bash
+docker build -t sentencepiece -f contrib/docker/Dockerfile .
+```
+
+[Run](https://docs.docker.com/reference/cli/docker/container/run/) the command line tools:
+
+```bash
+docker run --rm sentencepiece spm_encode --help
+```


taku910:
We will  not maintain Dockerfiles within our team. We are concerned that merging this will add to our maintenance and version upgrade tasks. Could you share your plan for sustainable maintenance? specifically, since the version is pinned to alpine:3.23, how do you plan to manage future updates?

If guaranteeing continuous maintenance is challenging, we suggest: 1) Moving this to a separate repository. 2) Placing it under a contrib directory to make sure that this file is not maintained.

kkew3:
Thank you so much for your review and suggestion! I have moved it to `contrib/docker/` directory and placed a disclaimer there. Does it look good?

I totally agree with your concern. In particular, `gperftools-dev` and `tcmalloc` packages are currently in `edge` branch and `testing` repository; they may be merged to `main` repository in the future, which may break the Dockerfile.

While I'm willing to maintain, it's indeed challenging to make a guarantee. And to be honest, I'm less experienced in maintaining large long-term project, and I really don't have a constructive plan for it (besides pulling from sentencepiece repo and building myself and see if it breaks from time to time, and probably configure a dependabot in my fork?). But I'd love to help to make sentencepiece easier to deploy and use. Would you give me advice on it? Thanks again.

kkew3:
Just to let you know that I've set up a dependabot (kkew3/sentencepiece_dockerfile@00a3bb429624b87872168c03e28fb4df08b5f9a7) and a cron job that periodically runs `spm_train` on a trivial corpus (kkew3/sentencepiece_dockerfile@2e59cbb80ba4f5244b6526bc12a0fa707393e416) at my fork for automatic continual maintenance of the dockerfile.

PR #1199, opened 2026-03-23
Add early return when nbest_samples is empty to prevent potential issues

### LINKED ISSUE #1198: SentencePieceProcessor::SampleEncodeAndScore(...) segfaults for a single quote with wor=true and include_best=true
**Describe the bug**
I’m seeing a native crash when calling SentencePieceProcessor::SampleEncodeAndScore(...) on a specific input/model combination.

**To Reproduce**
In python, the repro looks like this
```
import sentencepiece as spm

sp = spm.SentencePieceProcessor()
sp.load("10240-nmt_nfkc_cf.spm")
sp.sample_encode_and_score("'", num_samples=1, alpha=0.1, wor=True, include_best=True, out_type=str)
```

This also seems to be directly mapping to C++ test

```
TEST(SentencePieceProcessorTest, SampleEncodeAndScoreRegressionSingleQuote) {
  SentencePieceProcessor sp;

  const std::string model_path = "10240-nmt_nfkc_cf.spm";

  EXPECT_TRUE(sp.Load(model_path).ok());

  std::vector<std::pair<std::vector<std::string>, float>> pieces;
  EXPECT_TRUE(sp.SampleEncodeAndScore("'", 1, 0.1f, true, true, &pieces).ok());

  ASSERT_EQ(1, pieces.size());
  ASSERT_FALSE(pieces[0].first.empty());
  EXPECT_TRUE(std::isfinite(pieces[0].second));
}
```
The test binary aborts with `Segmentation fault: 11`.

**Expected behavior**
The method should either:
* return a valid sampled result, or
* fail gracefully with an error status

It should not crash the process.

**Additional context**
Unfortunately, I cannot share the model itself. I can provide more assistance in tracking the bug. 

I also observed that the `wor=true`, `include_best=false` path returns an error status rather than crashing, with:
`SampleEncodeAndScore` returns empty result.
So the crash seems tied specifically to the `include_best=true` branch when the sampled result is empty.

jtrmal:
Prevents accessing `nbest_samples.back()` when the container is empty, which would cause undefined behavior.

Fixes #1198

### DIFF
--- a/src/unigram_model.cc
+++ b/src/unigram_model.cc
@@ -791,6 +791,10 @@ NBestEncodeResult Model::SampleEncodeAndScore(absl::string_view normalized,
         nbest_samples.pop_back();
       }
     }
+    if (nbest_samples.empty()) {
+      return results;
+    }
+
     // We use the perturbed score of the k+1th element to calculate the
     // inclusion probability.
     const double kappa = static_cast<double>(nbest_samples.back().second);


jtrmal:
Looks like this wasn't a sufficient fix

PR #1207, opened 2026-04-02
Pass explicit key lengths in BuildTrie and add bounds checks to piece accessors

Aviral2642:
This patch addresses two hardening issues:

**1. BuildTrie passes nullptr for key lengths to trie_->build()**

`unigram_model.cc:629` passes `nullptr` as the lengths parameter to `Darts::DoubleArray::build()`, causing it to use `strlen()` to determine key lengths. This is the same pattern that was fixed for `PrefixMatcher` in commit d856b67 (CVE-2026-1260). Piece strings from the protobuf are not guaranteed to be null-terminated at the expected boundary, so explicit lengths should be passed.

**2. Piece accessor functions rely on GOOGLE_DCHECK for bounds checking**

All 13 piece accessor functions in `model_interface.h` (`IdToPiece`, `GetScore`, `IsUnused`, etc.) pass piece IDs directly to `model_proto_->pieces(id)` without bounds validation. The only bounds check is protobuf's `GOOGLE_DCHECK`, which is compiled away in release builds (`-DNDEBUG`). This patch adds `CHECK_GE`/`CHECK_LT` guards that remain active in all build configurations.

All existing tests pass.

### DIFF
--- a/src/model_interface.h
+++ b/src/model_interface.h
@@ -136,6 +136,8 @@ class ModelInterface {
   // Returns the string representation of vocab with `id`.
   // id must be 0 <= id < GetPieceSize().
   virtual const std::string &IdToPiece(int id) const {
+    CHECK_GE(id, 0);
+    CHECK_LT(id, model_proto_->pieces_size());
     return model_proto_->pieces(id).piece();
   }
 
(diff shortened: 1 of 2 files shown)


src/model_interface.h
```diff
@@ -136,6 +136,8 @@ class ModelInterface {
   // Returns the string representation of vocab with `id`.
   // id must be 0 <= id < GetPieceSize().
   virtual const std::string &IdToPiece(int id) const {
+    CHECK_GE(id, 0);
```
taku910: Please remove CHECK from *.h.  We don't allow to use CHECK in prod code (DCHECK) could be okay though.

Aviral2642: Done — replaced all CHECK with DCHECK in the header. Thanks for the review!

Aviral2642 (commented):
Done 
 replaced all CHECK with DCHECK in the header. Thanks for the review!

PR #1215, opened 2026-04-15
Add runtime bounds checks for piece ID accessors

sharadboni:
## Summary

The 13 piece ID accessor functions in `ModelInterface` (`src/model_interface.h`) had their `DCHECK` bounds checks commented out since commit bd8cf19. Without validation, a crafted `.model` file with a malicious trie can return out-of-range piece IDs, causing **out-of-bounds reads** on the `model_proto_->pieces()` array.

**Affected functions (7 virtual + 6 inlined):**
- `IdToPiece`, `GetScore`, `IsUnknown`, `IsControl`, `IsUnused`, `IsUserDefined`, `IsByte`
- `GetScoreInlined`, `IsUnknownInlined`, `IsControlInlined`, `IsUnusedInlined`, `IsUserDefinedInlined`, `IsByteInlined`

## Fix

Replace the commented-out `DCHECK` macros with runtime bounds checks (`if (id < 0 || id >= model_proto_->pieces_size())`) that return safe defaults for out-of-range IDs:

- `const std::string&` functions: return a static empty string
- `float` functions: return `0.0f`
- `bool` functions: return `false`

This prevents OOB memory access when processing untrusted model files while maintaining the same API behavior for valid inputs.

## Reproduction

Load a crafted `.model` file where the serialized trie contains node IDs exceeding `pieces_size()`. During tokenization, the model's `Encode()` path calls `IdToPiece()` / `GetScore()` / `IsControl()` etc. with the trie-returned IDs, which access `model_proto_->pieces(id)` without bounds checking, reading out-of-bounds memory.

### DIFF
--- a/src/model_interface.h
+++ b/src/model_interface.h
@@ -136,8 +136,10 @@ class ModelInterface {
   // Returns the string representation of vocab with `id`.
   // id must be 0 <= id < GetPieceSize().
   virtual const std::string &IdToPiece(int id) const {
-    // DCHECK_GE(id, 0);
-    // DCHECK_LT(id, model_proto_->pieces_size());
+    if (id < 0 || id >= model_proto_->pieces_size()) {
+      static const std::string kEmpty;
+      return kEmpty;
+    }
     return model_proto_->pieces(id).piece();
   }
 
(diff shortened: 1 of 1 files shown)


sharadboni:
@taku910 Could you review this security fix? It restores bounds checks on piece ID accessor functions in model_interface.h that were removed when DCHECKs were commented out, preventing OOB reads from crafted model files.

taku910:
In C++, it is a convention for the user to handle boundary checks because they can impact performance. The STL typically does not perform boundary checks on no-debug build.
https://en.cppreference.com/cpp/language/Zero-overhead_principle

Could you clarify the use cases for this check when dealing with untrusted models? If an ID is out of bounds, that's just a straight-up access violation rather than a matter of the model's reliability.

taku910:
We are afraid we cannot merge this PR for the following reasons:

1. Significant Spec Change: Returning a default value on a range error would be a major change to the specification. Currently, we follow STL and protobuf conventions, where range errors result in undefined behavior. Changing this would have too many side effects.
2. C++ Conventions: In C++, it is standard practice for the user (the caller) to perform range checks.
3. Security Scope: This is not considered a security vulnerability. It would be a serious issue if undefined behavior occurred even after the user performed a range check. Please let us know if you have a PoC demonstrating such a case.

FYI:  Python module does perform the range check and raises an exception if there is an error.

PR #1218, opened 2026-04-24
Fix segfault in SampleEncodeAndScore when wor=true and include_best=true with single-path input

### LINKED ISSUE #1198: SentencePieceProcessor::SampleEncodeAndScore(...) segfaults for a single quote with wor=true and include_best=true

anishesg:
In `unigram_model.cc`, `Model::SampleEncodeAndScore` segfaults when called with `wor=true` and `include_best=true` on inputs that produce only one unique Viterbi path (e.g. a single quote with certain models).

The root cause is in the `wor=true` branch: `NBest(samples + 1, ...)` is called to draw one extra sample for computing the Gumbel threshold (kappa), but when the input has only a single unique path, after the best path is removed from `nbest_samples`, the vector becomes empty. The code then unconditionally calls `nbest_samples.back()` on line 798, which is undefined behaviour and crashes.

The fix adds an emptiness check on `nbest_samples` immediately after the best-path removal block. If no additional samples remain, the function returns early with just the best-path result that was already pushed into `results`, mirroring the graceful handling already present in the `include_best=false` path.

Fixes #1198

### DIFF
--- a/src/unigram_model.cc
+++ b/src/unigram_model.cc
@@ -793,6 +793,13 @@ NBestEncodeResult Model::SampleEncodeAndScore(absl::string_view normalized,
         nbest_samples.pop_back();
       }
     }
+
+    if (nbest_samples.empty()) {
+      // No additional samples available (e.g. only one unique path exists).
+      // Return the best-path result that was already added above.
+      return results;
+    }
+
     // We use the perturbed score of the k+1th element to calculate the
     // inclusion probability.
     const double kappa = static_cast<double>(nbest_samples.back().second);
--- a/src/unigram_model_test.cc
+++ b/src/unigram_model_test.cc
@@ -612,6 +612,33 @@ TEST(UnigramModelTest, SampleEncodeAndScoreTest) {
   }
 }
 
+TEST(UnigramModelTest, SampleEncodeAndScoreSinglePathTest) {
+  // Regression test for segfault when wor=true and include_best=true on an
+  // input that has only a single unique Viterbi path (issue #1198).
+  ModelProto model_proto = MakeBaseModelProto();
+  // Only one piece matches "A", so there is exactly one segmentation path.
+  AddPiece(&model_proto, "A", 0.0);    // 3
+  AddPiece(&model_proto, "B", 0.0);    // 4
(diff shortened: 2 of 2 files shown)


taku910:
Thank you. Could you add the test case for this change?

anishesg:
Added `SampleEncodeAndScoreSinglePathTest` in `unigram_model_test.cc` — it builds a model where the input `"A"` has exactly one segmentation path, then calls `SampleEncodeAndScore` with `wor=true, include_best=true` for several `samples` values. Verifies no crash and that the result is exactly the best-path entry.

PR #1224, opened 2026-05-01
Validate darts-clone trie offsets at model load (v2)

TristanInSec:
This is a revised version of #1216, fixing the test regression caused by pre-compiled NFKC normalization maps.

**Root cause of the test failure:** Leaf/value units in the darts-clone double array store an integer value in their `unit_` field, not a child pointer. The `offset()` accessor on these units returns meaningless bits extracted from the stored value. The original `validate()` checked every unit's offset indiscriminately, so it rejected valid pre-compiled maps whose leaf values happened to decode to large offsets.

**Fix:** Skip units where `label() > 0xFF` (MSB set), which identifies leaf/value units. This matches the existing pattern already used in `open()` (line 380), where the same guard appears:

```cpp
if (units[i].label() <= 0xFF && units[i].offset() >= size) {
```

**What this PR does:**
- Adds `validate()` to `DoubleArrayImpl` in `third_party/darts_clone/darts.h`
- Calls it in `Normalizer::Init()` after `set_array()` to reject corrupted model files
- Correctly skips leaf/value units, preventing false positives on pre-compiled maps

The `GetPrecompiledCharsMapTest` test case now passes because leaf units in the NFKC maps are no longer misinterpreted as having out-of-bounds child pointers.

### DIFF
--- a/src/normalizer.cc
+++ b/src/normalizer.cc
@@ -62,6 +62,12 @@ void Normalizer::Init() {
     // but the number of double array units.
     trie_->set_array(const_cast<char *>(trie_blob.data()),
                      trie_blob.size() / trie_->unit_size());
+
+    if (!trie_->validate()) {
+      status_ = util::InternalError(
+          "Trie data contains out-of-bounds node references.");
+      return;
+    }
   }
 }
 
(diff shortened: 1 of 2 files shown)


taku910:
Thank you for the contribution.

Issue #1149, opened 2025-09-21
Use official abseil-cpp


taku910:
We are currently using a compatibility subset of the Abseil code, but we would like to replace it with the official package. This is expected to reduce maintenance costs and also improve performance. For example, the benefits of using absl::flat_hash_map are significant

Varshith2403315:
Hi, I’m a first-time contributor and I’d like to help with this migration.

From what I understand, the project currently uses a compatibility subset of Abseil, and the goal is to replace it with the official abseil-cpp dependency (e.g., using absl::flat_hash_map directly) to reduce maintenance and improve performance.

To keep the change scoped, I’d like to start by:

Identifying where the current Abseil compatibility code is used
Replacing one well-defined component (for example, a specific map/set usage) with the official absl equivalent
Updating build configuration as needed for that part

Does that sound reasonable, or is there a specific sub-area you’d prefer me to work on?

taku910:
Thank you.

Currently, we can build SentencePiece with the official Abseil library using `-DSPM_ABSL_PROVIDER=module`. As a result, the migration is almost finished, and the C++ binary and libraries are working as expected.

One launch-gating issue is the linkage of the Python module. Some functions and methods are not being linked correctly in the Python module.

```
ImportError: sentencepiece/python/build/lib.linux-x86_64-cpython-310/sentencepiece/_sentencepiece.cpython-310-x86_64-linux-gnu.so: undefined symbol: _ZN4absl12lts_2025081418container_internal19GetRefForEmptyClassERNS1_12CommonFieldsE
```

jovan2009:
Hi, I just tried to build on Windows from latest source code using `-DSPM_ABSL_PROVIDER=module`. I also get linking errors when I try to build Python wheel.

Also I'd like to mention that this symlink creation step doesn't work on Windows unless you are running the building with administrative privileges: 

https://github.com/google/sentencepiece/blob/b3235ab5b40c6e6416ca3dbe91550c3fbb830baa/CMakeLists.txt#L199-L201

taku910:
I could finally build a Python module with the official Abseil. In addition, we can make a smaller *.so file by explicitly removing the unnecessary symbols. I will send a PR soon.

taku910:
https://github.com/google/sentencepiece/pull/1236

taku910:
Finished the migration. Thank you.

PR #1234, opened 2026-05-10
Add parallel encoding support for handling super-long text

Senthil455:
This PR adds parallel encoding capabilities to SentencePiece to improve performance when tokenizing very long texts. The implementation splits long input text into chunks, processes them in parallel using multiple threads, and combines the results.

### Key Changes:

**New API Methods:**
- `EncodeParallel()` - Multi-threaded versions of existing encode methods
- `EncodeAsPiecesParallel()` - Convenient wrapper for piece output
- `EncodeAsIdsParallel()` - Convenient wrapper for ID output  
- `EncodeAsImmutableProtoParallel()` - Proto-based API with parallel processing

**Implementation Details:**
- **UTF-8 Safe Splitting:** Text is split at character boundaries to avoid breaking multi-byte sequences
- **Normalization First:** Entire input is normalized once, then chunks are encoded in parallel
- **Thread Management:** Uses `std::async` for parallel execution with configurable thread count
- **Fallback Logic:** Automatically uses single-threaded encoding for short texts (< 10KB) or when `num_threads = 1`
- **Offset Handling:** Maintains correct byte-offset mappings between original and tokenized text

**Usage Example:**
```cpp
SentencePieceProcessor sp;
sp.Load("model.model");

// Encode with 8 threads for better performance on long texts
std::vector<std::string> pieces;
sp.EncodeParallel("very long text content...", &pieces, 8);
```

**Benefits:**
- Significant performance improvements for tokenizing super-long contexts
- Maintains full API compatibility and correctness
- Memory efficient - normalizes once, processes chunks
- Configurable thread count with sensible defaults

**Testing:**
- All existing functionality remains unchanged
- New parallel methods produce identical results to single-threaded encoding
- UTF-8 boundary handling ensures correctness across different languages

This feature is particularly useful for applications dealing with large documents, long-form content, or batch processing of extensive text corpora where tokenization performance is critical.

### DIFF
--- a/src/sentencepiece_processor.cc
+++ b/src/sentencepiece_processor.cc
@@ -17,10 +17,12 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <future>
 #include <iterator>
 #include <map>
 #include <memory>
 #include <set>
+#include <thread>
 #include <utility>
 #include <vector>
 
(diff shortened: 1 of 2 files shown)


taku910:
I'm sorry, but we won't be merging this PR.

We already have internal code for parallel encoding that is more refined in several ways. However, it hasn't been merged yet because its threading implementation depends on the full Abseil suite. We are currently in the process of migrating the codebase to support the full set of Abseil libraries.
https://github.com/google/sentencepiece/issues/1149

Here is a high-level summary of the differences:
- std::async is not allowed to be used in production code.
- Dependency Injection: Our design allows for the ThreadPool to be injected externally to reduce the cost of thread creation.
- Tokenization Consistency: It includes workaround code for cases where tokenization consistency cannot be maintained at sentence boundaries.

PR #1243, opened 2026-05-18
uses abseil status

taku910:
Uses abseil-status instead of the internal status library.


taku910:
decided to hold off on merging this pull request for now, as it introduces a significant number of side effects.

PR #1260, opened 2026-06-03
update cpp version to 20

taku910:
Update CPP version from 17 to 20.


taku910:
Seems that win build requires Cpp17 to build the old protobuf..

PR #1261, opened 2026-06-05
Support parallel encoding

### LINKED ISSUE #1229: Support parallel encoding
Support parallel encoding to handle super-long text.

Parallel encoding splits long text into multiple chunks and encodes them in parallel. This would be useful for tokenizing super-long context.

Senthil455:
Support parallel encoding to handle super-long text. When input exceeds 10KB, the text is split into chunks and encoded concurrently using `std::async`, then the results are stitched back together.
### `src/sentencepiece_processor.cc`
- **`EncodeParallel(input, pieces, num_threads)`** — new overload returning `vector<string>`
- **`EncodeParallel(input, ids, num_threads)`** — new overload returning `vector<int>`
- **`EncodeParallel(input, spt, num_threads)`** — new overload returning `SentencePieceText` (core implementation)
  - Falls back to serial `Encode()` for inputs <10KB or when `num_threads=1`
  - Normalizes the entire input first, then splits into chunks via `SplitInputIntoChunks`
  - Encodes each chunk with `std::async` and combines results
  - Calls `PopulateSentencePieceText` to build the final output with correct byte offsets
- **`SplitInputIntoChunks(input, num_chunks)`** — new helper that splits text at UTF-8 character boundaries (avoids splitting multi-byte sequences)
- Added `#include <future>` and `#include <thread>`
### `src/sentencepiece_processor.h`
- Added `EncodeParallel` virtual methods (3 overloads) to the public API
- Added convenience wrappers:
  - `EncodeAsPiecesParallel(input, num_threads)`
  - `EncodeAsIdsParallel(input, num_threads)`
  - `EncodeAsImmutableProtoParallel(input, num_threads)`
- Added `SplitInputIntoChunks` as a private helper
### `src/sentencepiece_processor_test.cc`
Added 5 test cases:
| Test | Description |
|------|-------------|
| `EncodeParallelShortInputTest` | Input <10KB falls back to serial; tests all 3 overloads + convenience wrappers |
| `EncodeParallelSingleThreadTest` | `num_threads=1` matches serial `Encode()` |
| `EncodeParallelInvalidNumThreadsTest` | `num_threads=0` returns error |
| `EncodeParallelConsistentWithSerialTest` | 15KB input produces identical results from parallel (4 threads) vs serial |
| `EncodeParallelMultiByteUtf8Test` | 12KB input with 3-byte UTF-8 chars (あ) across chunk boundaries — parallel matches serial |
Also added `CharByCharModel` test stub that tokenizes character-by-character without input verification, enabling parallel-mode testing.
## Implementation Details
- Chunk boundaries are adjusted backward to avoid splitting UTF-8 continuation bytes (`string_util::IsTrailByte`)
- Uses `std::async(std::launch::async, ...)` for thread-safe parallel execution
- Normalization is done once upfront, then chunks work on the normalized text
- Final `PopulateSentencePieceText` maps byte offsets back to original (unnormalized) input
Fixes #1229

### DIFF
--- a/src/sentencepiece_processor.cc
+++ b/src/sentencepiece_processor.cc
@@ -715,13 +715,11 @@ util::Status SentencePieceProcessor::EncodeParallel(absl::string_view input,
 
   // Wait for all encodings to complete and combine results
   EncodeResult combined_result;
-  size_t normalized_offset = 0;
   for (size_t i = 0; i < futures.size(); ++i) {
     const auto chunk_result = futures[i].get();
     for (const auto &piece : chunk_result) {
       combined_result.emplace_back(piece.first, piece.second);
     }
-    normalized_offset += normalized_chunks[i].size();
   }
 
   // Populate the final SentencePieceText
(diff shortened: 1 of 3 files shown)


taku910:
ParallelEncode has already been implemented.  See the head.

There are some comments on the parallel encoding.
- Simply splitting the input is insufficient. If split at token boundaries, the tokenization result of the divided text may differ from the result of tokenizing the entire text all at once.
- Passing the number of threads lacks extensibility and prevents coordination with other threads. Therefore, the implementation now passes a ThreadPool. (Internally, passing num_threads has been removed.)

PR #1267, opened 2026-06-09
bound commonPrefixSearch result count to buffer size in normalizer

sahvx655-wq:
**Stack over-read in the normalizer trie lookups**

While running the normaliser under AddressSanitizer against a model whose `precompiled_charsmap` I had crafted by hand, ASAN flagged a stack-buffer-overflow read in `NormalizePrefix`, at the loop that scans `trie_results`. The on-stack array holds `kMaxTrieResultsSize` (32) entries, but `Darts::DoubleArray::commonPrefixSearch` returns the total number of prefix matches and only writes up to the buffer capacity (documented in darts.h), so the loop walks off the end of the array whenever a query position has more than 32 shared-prefix rules. The build path is safe because `CompileCharsMap` rejects maps that exceed that count, yet a `precompiled_charsmap` read straight out of a model proto never goes through that check, so the invariant the comment relies on is not actually present for loaded models. `PrefixMatcher::PrefixMatch` has the same shape with a 64-entry buffer fed from the model's user-defined symbols and over-reads identically once there are more than 64 nested symbols.

The root cause is treating the search return value as the number of entries written rather than a possibly larger match total. The fix caps the count to the buffer capacity at both sites before it is used as a loop bound. Left unfixed this is an out-of-bounds read reachable from any untrusted model (via `spm_normalize` or `SentencePieceProcessor::Load`), leaking adjacent stack into the selected rule and potentially crashing the process. I reproduced both over-reads under ASAN using tries of 40 and 70 nested keys; the added regression tests exercise both paths and the full suite passes clean afterwards.

### DIFF
--- a/src/normalizer.cc
+++ b/src/normalizer.cc
@@ -208,9 +208,16 @@ std::pair<absl::string_view, int> Normalizer::NormalizePrefix(
     Darts::DoubleArray::result_pair_type
         trie_results[Normalizer::kMaxTrieResultsSize];
 
-    const size_t num_nodes = trie_->commonPrefixSearch(
-        input.data(), trie_results, Normalizer::kMaxTrieResultsSize,
-        input.size());
+    // commonPrefixSearch returns the total number of matches, which can be
+    // larger than the buffer size (see darts.h). A precompiled charsmap loaded
+    // from an untrusted model is not subject to the build-time
+    // kMaxTrieResultsSize check, so cap the count to the buffer capacity before
+    // indexing trie_results.
+    const size_t num_nodes = std::min<size_t>(
+        trie_->commonPrefixSearch(input.data(), trie_results,
+                                  Normalizer::kMaxTrieResultsSize,
+                                  input.size()),
+        Normalizer::kMaxTrieResultsSize);
 
     // Finds the longest rule.
     for (size_t k = 0; k < num_nodes; ++k) {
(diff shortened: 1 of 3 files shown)


taku910:
Thank you. The unittest fails on PowerPC, powerpc64, and s390x (likely big-endian environments).

```
[ RUN      ] NormalizerTest.ManySharedPrefixesTest
[       NG ] src/normalizer_test.cc:438: failed: normalizer.status().ok()
Error: Process completed with exit code 255.
```

sahvx655-wq:
Good catch. The over-read fix itself is endian-neutral, the regression test wasn't. It hand-assembled the precompiled_charsmap as <uint32 size><trie><normalized> in the host's native byte order, which only matches the on-disk little-endian layout on little-endian builds. On big-endian, DecodePrecompiledCharsMap byte-swaps that size header expecting little-endian, so it read a bogus trie size, tripped the size/divisibility guard, and left status() not ok. That's the assert you saw at normalizer_test.cc:438.

I've reworked the test to build the blob through Normalizer::EncodePrecompiledCharsMap, the same path the builder and real model protos use, so the header and trie units get swapped to little-endian on encode and back on decode. It round trips cleanly on both endiannesses now, and the full suite still passes locally. The PrefixMatcher test was unaffected since it builds its trie in memory and never serialises.

PR #1272, opened 2026-06-13
reject out-of-bounds root unit in DoubleArray::validate

sahvx655-wq:
Loading a model whose `normalizer_spec.precompiled_charsmap` is attacker controlled can read out of bounds during normalisation. The charsmap is a darts double-array deserialised straight from the proto, and `Normalizer::Init` guards it with `Darts::DoubleArray::validate()`. Whilst tracing an ASAN `BUS` fault in `NormalizePrefix` (`commonPrefixSearch` at normalizer.cc:218) I found that `validate()` skips any unit whose MSB label bit is set, treating it as a leaf/value unit. The root unit (index 0) is special: `commonPrefixSearch()` and `traverse()` always follow its `offset()` before they inspect any label, so a root unit with the MSB set and a large offset field passes `validate()` and the very first transition indexes far outside the array. The model loads cleanly and the fault only surfaces on the first `Encode`/`Normalize`, or via `spm_normalize --decompile` (`DecompileCharsMap` then `traverse`). Left unpatched this is an attacker controlled out-of-bounds read reachable from any untrusted `.model` file.

`open()` already rejects a malformed root the same way (label must be `\0`, no leaf bit, offset in range); `validate()` simply never carried that check across, so both `Init` and `DecompileCharsMap` inherit the gap. I added the root check inside `validate()` using the same in-bounds rule the loop applies to every other node, which fixes both call sites with one guard. I verified it with a fuzzer that mutates a compiled charsmap and then normalises: the `BUS` fault reproduces on the unpatched tree and is gone afterwards, every embedded charsmap (`nfkc`, `nmt_nfkc`, `nfkc_cf`, ...) still loads and encodes, and the existing normalizer and builder tests stay green. `NormalizerTest.MalformedRootUnitTest` corrupts the root unit and asserts the model is rejected; it fails before the change and passes after.

### DIFF
--- a/src/normalizer.h
+++ b/src/normalizer.h
@@ -99,6 +99,7 @@ class Normalizer {
  private:
   FRIEND_TEST(NormalizerTest, EncodeDecodePrecompiledCharsMapTest);
   FRIEND_TEST(NormalizerTest, ManySharedPrefixesTest);
+  FRIEND_TEST(NormalizerTest, MalformedRootUnitTest);
 
   void Init();
 
--- a/src/normalizer_test.cc
+++ b/src/normalizer_test.cc
@@ -14,6 +14,8 @@
 
 #include "normalizer.h"
 
+#include <cstdint>
+#include <cstring>
 #include <set>
 #include <string>
 #include <vector>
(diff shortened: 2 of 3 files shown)


taku910:
Thank you for the PR.

As we mentioned in the issue, we've merged internal changes which fix various security risks, including the Dart issues (d685ef3).

We think this fix already covers this PR. Let's reopen the issue if you find any problems.

PR #1296, opened 2026-07-29
Integrate the latest Protobuf lite

taku910:
Use the latest protobuf lite


.github/workflows/cifuzz.yml
```diff
@@ -10,20 +10,20 @@ jobs:
...
       id: build
-      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@master
+      uses: google/oss-fuzz/infra/cifuzz/actions/build_fuzzers@c73e4ccf50c0634081e3255d40858e79a7f30c26
       with:
         oss-fuzz-project-name: 'sentencepiece'
         dry-run: false
         language: c++
     - name: Run Fuzzers
-      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@master
+      uses: google/oss-fuzz/infra/cifuzz/actions/run_fuzzers@c73e4ccf50c0634081e3255d40858e79a7f30c26
```
Marcono1234, line 19: Pinning `google/oss-fuzz` might not work as desired, see also https://github.com/google/oss-fuzz/issues/6836#issuecomment-5170184193 (and the other comments).

Dependabot will most likely not update it.

PR #1310, opened 2026-08-25
Introduce Bazel build system (Bzlmod)

taku910:
Add a bzlmod-based Bazel build alongside the existing CMake build. Covers core runtime, trainer, CLI binaries, standalone shared libraries (libsentencepiece.so, libsentencepiece_train.so), and full test suites. Includes documentation in doc/bazel.md and GitHub Actions CI workflow supporting Linux (x64/arm64), macOS (arm64/x64), and Windows (x64/arm64).

Closes #1280


BYVoid:
Happy to update https://registry.bazel.build/modules/sentencepiece when 0.2.3 is released

PR #1311, opened 2026-08-26
build: align the Bazel dependency versions with CMake

BYVoid:
`MODULE.bazel` and `CMakeLists.txt` pinned different versions of the shared dependencies. This takes all three from the Bazel Central Registry at the versions `CMakeLists.txt` already uses, which also removes both `git_override` entries:

| Dependency | CMake | Bazel (before) | Bazel (after) |
|---|---|---|---|
| abseil-cpp | `20260817.0` | `20250814.1` + `git_override` tag `20260817.0` | `20260817.0` |
| protobuf | `v36.0` | `35.0` | `36.0.bcr.1` |
| googletest | `v1.18.0` | `1.16.0` + `git_override` branch `main` | `1.18.0.bcr.1` |

Dropping the overrides means the build no longer clones abseil and googletest at fetch time, and pinning googletest to a release instead of tracking `main` makes the build reproducible.

protobuf is pinned to `36.0.bcr.1` rather than `36.0` because `36.0` records the wrong SHA256 for `protoc-36.0-osx-aarch_64.zip` in its `tool_integrity.bzl`, which breaks the fetch of the prebuilt `protoc` on macOS arm64. `36.0.bcr.1` patches that entry.

### Verified

macOS arm64, Bazel 9.1.1, with the default toolchain and no extra flags: `bazel build //...` (56 targets) and `bazel test //...` (22 of 22 tests pass).


taku910:
Thanks for the PR!
Regarding the `osx-aarch_64` checksum failure for protobuf 36.0: BCR has already yanked `36.0` and published `36.0.bcr.1` which includes a patch fixing the exact `tool_integrity.bzl` checksum for macOS arm64.
Could you update `MODULE.bazel` to use `36.0.bcr.1` for protobuf and update `MODULE.bazel.lock`?
```starlark
bazel_dep(name = "protobuf", version = "36.0.bcr.1", repo_name = "com_google_protobuf")

BYVoid:
Thanks, that was exactly it — updated.

`MODULE.bazel` now uses `36.0.bcr.1`, and `MODULE.bazel.lock` is regenerated. I confirmed on macOS arm64 (Bazel 9.1.1) that the prebuilt `protoc` fetch now succeeds with no extra flags: `bazel build //...` builds all 56 targets and `bazel test //...` passes 22 of 22.

The PR description is updated too, so it no longer describes the checksum failure as an open issue.

PR #1313, opened 2026-08-27
Take libsais from the Bazel Central Registry

BYVoid:
The Bazel build depends on the [`libsais` BCR module](https://registry.bazel.build/modules/libsais) instead of compiling `third_party/libsais/libsais.c`. The CMake build is unchanged and still compiles the vendored copy.

## `2.10.4.bcr.2` vs. `third_party/libsais`

Both are upstream `v2.10.4` — `third_party/libsais/libsais.h` declares `LIBSAIS_VERSION_STRING "2.10.4"`, and the module fetches the `v2.10.4` tarball. I diffed the vendored sources against that tarball:

| | result |
|---|---|
| `libsais.h` | identical once comments and formatting are normalised |
| `libsais.c` | identical once comments, formatting, include order and one `clang-format` line continuation are normalised |

So unlike `third_party/darts_clone`, the vendored `libsais` carries **no local patches** — it is upstream `v2.10.4` after this repo's IWYU/format pass. The module needs no patch to the library sources either, so it can be taken at a plain version with nothing overridden on this side. `.bcr.1` and `.bcr.2` are registry-side revisions of the same upstream source: `.bcr.1` adds the OpenMP toggle described below, and `.bcr.2` patches out the archive's `VERSION` file, without which MSVC resolves `#include <version>` to it on a case-insensitive file system (bazelbuild/bazel-central-registry#10291, reported from this PR's Windows jobs).

SentencePiece calls `libsais`, `libsais_omp`, `libsais_plcp`, `libsais_plcp_omp`, `libsais_lcp` and `libsais_lcp_omp`, all from `unigram_model_trainer.cc`. All six are in the module's `@libsais//:libsais` target (`uint8_t` input, `int32_t` suffix array); the `16`/`64`/`16x64` variants are unused.

## OpenMP

`2.10.4.bcr.1` and later expose an `--@libsais//:openmp` flag, the counterpart of libsais' own `LIBSAIS_USE_OPENMP` CMake option. It adds `-fopenmp` (or `/openmp`), and sets `LIBSAIS_OPENMP` as a `defines`, so the macro reaches dependents and `unigram_model_trainer.cc` compiles its `_omp` branch — the same way the CMake build's `add_compile_definitions(LIBSAIS_OPENMP)` reaches it.

It is **off by default**, which matches both libsais' own CMake default and the `//third_party:libsais` target this replaces, so this change does not alter what the Bazel build does today: it still takes the serial path. This PR does not turn it on, because `-fopenmp` is unsupported by Apple clang, so enabling it in `.bazelrc` would break the macOS Bazel job.

Checked with a toolchain that does support it (`--repo_env=CC=/opt/homebrew/opt/llvm/bin/clang`), building `-c opt` with and without the flag and timing the seed-extraction phase from the trainer's own log over an 8.7 MB corpus, 6 runs each, interleaved:

| | seed extraction | trainer log line |
|---|---|---|
| default | 0.446 s | `Extracting seed sentencepieces (libsais)` |
| `--@libsais//:openmp` | 0.396 s | `Extracting seed sentencepieces (libsais_omp, 8 threads)` |

So the flag works end to end and is worth roughly 1.13x here — suffix array construction does not parallelise dramatically at this corpus size. Mentioned only to show the path is wired correctly; the default is unchanged.

## Optimisation flags

`//third_party:libsais` set `copts = ["-O3"]`, which the module does not. Per `bazel aquery`:

| | before | after |
|---|---|---|
| `-c opt` | `-O3` | `-O2` |
| `-c fastbuild` | `-O3` | none |

I measured what that actually costs, on macOS/arm64 with Apple clang 21, two ways.

**Isolated.** A driver running the same `libsais` → `libsais_plcp` → `libsais_lcp` sequence as `ExtractSeedSentencePieces`, over an 8.7 MB corpus, with only `libsais.c`'s optimisation level varied (driver fixed at `-O2`; median of 24 runs):

| `libsais.c` | per pass |
|---|---|
| `-O0` | 0.368 s |
| `-O2` | 0.146 s |
| `-O3` | 0.147 s |

**End to end.** `spm_train --model_type=unigram --vocab_size=8000` over the same corpus, timing the seed-extraction phase from the trainer's log lines (6 runs each, interleaved order):

| build | seed extraction |
|---|---|
| `-c opt`, module as-is (`-O2`) | 0.439 s |
| `-c opt` + `--per_file_copt=…libsais…@-O3` | 0.438 s |

`-O3` buys nothing over `-O2` for this code — the two agree to within 1% in both measurements — so dropping the `copts` costs no measurable performance in `-c opt`. I originally assumed this would be the one real regression in the change; it measures as a non-issue, so I am not proposing a `.bazelrc` workaround unless you want one.

The `fastbuild` row is the larger change: libsais goes from `-O3` to unoptimised, which the isolated numbers put at ~2.5x slower for seed extraction. That only affects training run from a `fastbuild` binary, which was already an unoptimised build of everything else in the trainer.

Two caveats. This is one machine and one compiler; GCC on Linux may weigh `-O3` differently. And total `spm_train` wall time did shift by ~15% between the two binaries, but that difference sits entirely in the EM iterations, which do not call libsais and whose objects are byte-identical between the two builds — `aquery` confirms exactly one compile action changes. It is binary layout, not codegen, which is why the numbers above time the seed-extraction phase specifically rather than the whole run.

## Why the module rather than the vendored copy

**Upgrades become a version bump.** libsais is a single 400 KB `.c` file. Moving it forward today means re-dropping that file and re-running the format pass over it; with the module it is one line in `MODULE.bazel`.

**Known provenance.** The module pins the upstream `v2.10.4` tarball by SHA-256. The vendored copy is a reformatted snapshot whose correspondence to a specific upstream release can only be established by diffing, as above.

**Visible in dependency tooling.** A `bazel_dep` with a version appears in `MODULE.bazel.lock`, `bazel mod graph`, and any SBOM generated from them. A checked-in `.c` file does not.

The vendored copy is not deleted: the CMake build still compiles it, so this reduces the number of builds depending on the unlabelled snapshot from two to one rather than to zero.

## Include path

The module exports the header as `libsais.h` (`includes = ["include"]`), so the one `#include "third_party/libsais/libsais.h"` in `unigram_model_trainer.cc` loses the prefix. `src/CMakeLists.txt` gains `third_party/libsais` on its include path so the CMake build resolves the same spelling against the vendored copy. `libsais.h` is not reachable from any installed header, so the `target_include_directories(... PUBLIC ...)` blocks need no change.

## Verified

macOS/arm64:

- `bazel test //...` — 22 of 22 tests pass
- `cmake --build` — `sentencepiece_train-static` builds, compiling both `unigram_model_trainer.cc` and the vendored `libsais.c`

## Note

This touches `third_party/BUILD.bazel`, as does #1312, so whichever lands second will need a trivial conflict resolution there. The two changes are otherwise independent.


taku910:
Thanks for working on this! The Bazel Bzlmod migration looks great and tests pass cleanly.

One minor suggestion for consistency with #1312:
In #1312, `$<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party/darts_clone>` was added to `target_include_directories` in `src/CMakeLists.txt`.

To keep the CMake targets consistent and adhere to modern target-centric CMake practices, could we also add `third_party/libsais` to `target_include_directories` for `sentencepiece_train-static` (and `sentencepiece_train` under `SPM_ENABLE_SHARED`)?

```diff
diff --git a/src/CMakeLists.txt b/src/CMakeLists.txt
index ca29f81..xxxxxxx 100644
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -217,6 +217,7 @@ target_include_directories(sentencepiece_train-static PUBLIC
   $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party>
   $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party/darts_clone>
+  $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party/libsais>
   $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
 )
 
@@ -249,6 +250,7 @@ if (SPM_ENABLE_SHARED)
     $<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}>
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party>
     $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party/darts_clone>
+    $<BUILD_INTERFACE:${PROJECT_SOURCE_DIR}/third_party/libsais>
     $<INSTALL_INTERFACE:${CMAKE_INSTALL_INCLUDEDIR}>
   )

BYVoid:
Thanks for the review. I would like to push back gently on this one, and I have not changed the CMake yet — I would rather agree on the scope with you first.

The consistency argument holds for #1312 but not, I think, here, because the two headers sit at different levels:

| | included from |
|---|---|
| `darts.h` | `model_interface.h`, `normalizer.h`, `unigram_model.h` — headers |
| `libsais.h` | `unigram_model_trainer.cc` only — no header |

`PUBLIC` is `PRIVATE` plus `INTERFACE`. The `INTERFACE` half exists so that a target which includes one of *our* headers can resolve what that header pulls in. For `darts.h` that is real: anything in-tree including `normalizer.h` transitively includes `darts.h` and needs the directory, so `PUBLIC` in #1312 is correct. For `libsais.h` there is nothing to carry — it is reached only from a `.cc`, so the `INTERFACE` half would advertise a requirement that no dependent has.

It would not break anything, so if you prefer the uniform spelling I will just add it. But my suggestion is the other direction: replace the directory-scope `include_directories()` I currently have — which is the older style your comment is steering away from — with a target-scope `PRIVATE` entry on the targets that actually compile `unigram_model_trainer.cc`. That gets the target-centric form you are asking for without exporting a path no dependent uses, and keeps one fewer directory on the search path for everything that links the trainer.

Which would you like? I am happy either way; I just did not want to widen the scope silently.

Separately, on the red Windows jobs: those are not caused by this change as such. The libsais archive has a 7-byte `VERSION` file at its root, Bazel hands a dependency's repository root to dependents as a quote-include, and since MSVC has no `-iquote` it becomes a plain `/I`. On a case-insensitive file system `#include <version>` then resolves to that file instead of the C++ standard header. bazelbuild/bazel-central-registry#10291 publishes `2.10.4.bcr.2`, which patches the file away and adds a regression test; once it lands I will bump this PR to that version and the Windows jobs should go green.

PR #1322, opened 2026-09-03
build: make VERSION.txt the single source of truth for the project version

BYVoid:
## Problem

The project version lived in four hand-written places, and only one of them was kept in sync by tooling:

| File | Kept in sync by |
|---|---|
| `VERSION.txt` | source of truth for CMake and Bazel |
| `MODULE.bazel` | `.github/scripts/update_dependencies.py`, when it happens to run |
| `python/pyproject.toml` | nothing |
| `python/src/sentencepiece/_version.py` | nothing |

Nothing failed when they disagreed, so a release could ship a wheel whose `sentencepiece.__version__` did not match the C++ library it wrapped. Two documentation examples that users copy, the `bazel_dep` snippet in `doc/bazel.md` and the `FetchContent` `GIT_TAG` in `doc/cpp.md`, could also silently go stale.

## Approach

Derive the version wherever a build system can read a file, and test the one copy that cannot be derived. Bazel requires a literal in `module(version = ...)`, so `MODULE.bazel` must stay hand-written; everything else now follows from `VERSION.txt`.

### Python no longer duplicates the version

`python/setup.py` reads `VERSION.txt` and generates `src/sentencepiece/_version.py` from it. It looks in `../VERSION.txt` for a repository checkout and `./sentencepiece/VERSION.txt` for an sdist built by `build_sdist.py`. `pyproject.toml` declares `dynamic = ["version"]` instead of pinning a literal, and `_version.py` is no longer tracked in git since it is now a build artefact.

### One test, run by both build systems

`src/version_test.cc` asserts that `SPM_BUILD_SYSTEM_MODULE_VERSION`, the version the active build system declares, equals `VERSION` from the generated `config.h`. Each build system supplies its own value, so neither reads the other's files:

- **Bazel** passes `module_version()` from `MODULE.bazel` through `local_defines` in `//src:version_test`. This is the copy that can genuinely drift, and this is what catches it.
- **CMake** passes `PROJECT_VERSION` from `project(VERSION ...)`, with `version_test.cc` joining `SPM_TEST_SRCS`.

On the CMake side this is a structural invariant rather than a drift check, since CMake derives both `PROJECT_VERSION` and `config.h` from `VERSION.txt`. It guards against a stale or shadowing `config.h`. A second CMake test, `sentencepiece_version_test`, runs `spm_encode --version` and requires `sentencepiece ${SPM_VERSION}`, covering the path from `VERSION.txt` through to the built binaries.

### A release checklist that CI enforces

`.github/scripts/check_version.py` compares `VERSION.txt` against `MODULE.bazel`, the `bazel_dep` example in `doc/bazel.md` and the `FetchContent` `GIT_TAG` in `doc/cpp.md`, and confirms `pyproject.toml` carries no static version. The `GIT_TAG` was not previously covered anywhere.

Its module docstring is the release documentation: which four files a version bump must edit, which values are derived and must not be touched, and which version mentions are historical feature-availability notes such as `v0.2.3+` in `doc/options.md` that must keep their original value. A new `version_check.yml` workflow runs the script and also asserts `setup.py --version` equals `VERSION.txt`.

### Pinned GitHub Actions

The zizmor scan rejects unpinned action references. Every `uses:` in `.github/workflows/` is now pinned to a commit hash with a version comment, matching the convention already used by `cmake.yml`, `bazel.yml` and `wheel.yml`. This covers the new workflow plus the pre-existing unpinned references in `update_dependencies.yml` and `stale.yml`, and the new checkout sets `persist-credentials: false`.

## Verification

Both build systems were configured, built and run locally, with negative tests that temporarily introduce a mismatch:

- `bazel test //src:version_test` passes. Setting `MODULE.bazel` to `0.2.2` makes it fail with `The build system declares version "0.2.2" but VERSION.txt contains "0.2.3"`.
- CMake `ctest` passes both `sentencepiece_test`, which now contains `VersionTest`, and `sentencepiece_version_test`.
- `check_version.py` passes, and each of its three targets was broken individually to confirm it is reported. Historical `v0.2.3+` mentions in the docs are correctly left alone.
- Simulating the sdist layout, where only `sentencepiece/VERSION.txt` exists, `setup.py` resolves `0.2.3` and generates `_version.py`.
- `setuptools.build_meta.prepare_metadata_for_build_wheel` yields `Version: 0.2.3`, confirming the dynamic version works on the PEP 517 path used by `python -m build` and cibuildwheel.
- `zizmor --persona=regular` reports no findings for the new workflow and no remaining `unpinned-uses` in the repository.

## Note for reviewers

Importing the package directly from `python/src` without ever running `setup.py` now fails, because `_version.py` is generated rather than committed. Every documented build path runs `setup.py` first, so this affects only ad-hoc use of the source tree.


tiancixiao:
Please resolve the conflicts.

PR #1323, opened 2026-09-05
reject out-of-range id in SentencePieceProcessor id accessors

sahvx655-wq:
The public id accessors on `SentencePieceProcessor` (`IdToPiece`, `GetScore`, `IsControl`, `IsUnknown`, `IsUnused`, `IsByte`) forward their `id` straight to the model layer, which lands on `model_proto_->pieces(id)` guarded only by `DCHECK`. In a release build (`NDEBUG`) that check is compiled out, so an `id` outside `[0, GetPieceSize())` walks off the piece array. I noticed the asymmetry while reading the accessors: `SafeIdToPiece` and `Decode(absl::Span<const int>)` both range-check the id against `GetPieceSize()`, but these six siblings do not, even though they take the same untrusted id surface (token ids arriving from a caller or a model server). Building with `-fsanitize=address`, `sp.GetScore(sp.GetPieceSize())` aborts with a SEGV in `ModelInterface::GetScore` (`model_interface.h`) as the stray `RepeatedPtrField` element pointer is dereferenced; without instrumentation it silently returns garbage or crashes. Left unfixed this is a memory-safety fault reachable from any code path that queries a model-supplied or client-supplied id.

The fix adds the same `[0, GetPieceSize())` guard the safe siblings already use, inside each accessor, returning that method's existing not-loaded default when the id is out of range. Keeping the bound in the accessor rather than expecting every caller to pre-validate matches `SafeIdToPiece` and stops the public API being driven past the end of the array. The added case in `EndToEndTest` exercises `-1`, `GetPieceSize()` and a far past-the-end id across all six accessors; it SEGVs on the current tree under ASan and passes with the guard, and the full `ctest` suite stays green.

### DIFF
--- a/src/sentencepiece_processor.cc
+++ b/src/sentencepiece_processor.cc
@@ -1039,6 +1039,9 @@ int SentencePieceProcessor::PieceToId(absl::string_view piece) const {
 const std::string& SentencePieceProcessor::IdToPiece(int id) const {
   static const std::string* kEmptyString = new std::string;
   RET_CHECK_OR_RETURN_DEFAULT(*kEmptyString);
+  if (id < 0 || id >= model_->GetPieceSize()) {
+    return *kEmptyString;
+  }
   return model_->IdToPiece(id);
 }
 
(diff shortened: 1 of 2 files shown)


taku910:
Currently, not performing out-of-range checks is by design. In C++, it is standard practice to omit range checks in accordance with the zero-overhead principle (e.g., std::vector, protobuf’s RepeatedFields).

However, this may change in the future. Recent findings indicate that such checks incur almost no overhead: https://security.googleblog.com/2024/11/retrofitting-spatial-safety-to-hundreds.html

If we decide to make changes, we will provide an announcement beforehand and carefully consider which version to implement them in.

PR #1329, opened 2026-09-11
fix(builder): validate leaf node offsets and safely slice values in DecompileCharsMap

jdymitarai:
### Summary

This PR hardens `Builder::DecompileCharsMap` against corrupted or crafted normalization rule charsmaps by:
1. **Validating leaf node value offsets during trie validation**: In `Normalizer::Init()`, `trie_->validate(normalized_.size())` is called to ensure that all leaf node value offsets do not exceed `normalized_.size()`. However, `Builder::DecompileCharsMap` called `trie.validate()` without passing `normalized.size()`, skipping leaf offset bounds checking before starting the recursive traversal. We now pass `normalized.size()` to `trie.validate(normalized.size())`.
2. **Safely slicing values**: When reading normalization values at `result` during trie traversal, `normalized.find('\0', result)` is used to locate the null delimiter and extract `normalized.substr(result, null_pos - result)` instead of constructing `absl::string_view(normalized.data() + result)` which relied on unbounded `strlen()`.
3. **Adding regression tests**: Added unit tests in both `src/builder_test.cc` and `src/invalid_model_test.cc` verifying that precompiled charsmaps with out-of-bounds leaf offsets are safely rejected.

### Verification
- Added test case in `BuilderTest.DecompileMalformedCharsMapTest` verifying that leaf units pointing at or beyond `normalized.size()` are rejected.
- Added test case in `SentencePieceProcessorTest.RejectModelWithOOBCharsmapValue` verifying that `Builder::DecompileCharsMap` also rejects invalid model charsmaps.


taku910:
This looks generally okay, but this new test is failing in a big-endian environment.

jdymitarai:
Thanks @taku910! In 0bd31e3, updated the test helper \make_blob\ in \uilder_test.cc\ to encode the trie size and units using \bsl::little_endian::Store32\. Because \DecodePrecompiledCharsMap\ expects little-endian wire format and byte-swaps on big-endian hosts, encoding in host-native byte order previously caused the size check to fail on s390x/powerpc. The test now matches the wire format and passes on both little-endian and big-endian architectures.

